PR #9, opened 2018-07-02
DebianRepoChecker: improve flexibility of Debian repo checking

ramcq:
Add support for .bz2 Packages files, and repos with an exact URL which are usually specified in sources.list with just a base plus the subdirectory where Packages.* can be found.

### DIFF
--- a/src/checkers/debianrepochecker.py
+++ b/src/checkers/debianrepochecker.py
@@ -40,7 +40,8 @@ import urllib
 from lib.externaldata import ExternalData, CheckerRegistry, Checker
 from lib import utils
 
-DEB_PACKAGES_URL = '{root}/dists/{dist}/{comp}/binary-{arch}/Packages'
+DEB_PACKAGES_DISTRO_URL = '{root}/dists/{dist}/{comp}/binary-{arch}/Packages'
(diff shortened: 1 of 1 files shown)


src/checkers/debianrepochecker.py
```diff
@@ -155,8 +156,12 @@ def _split_packages_from_page(self, packages_page):
         return packages
 
     def _get_package_from_url(self, package_name, root, dist, component, arch):
-        packages_url = DEB_PACKAGES_URL.format(root=root, dist=dist,
-                                               comp=component, arch=arch)
+        if dist[-1:] == '/' and component is None:
```
joaquimrocha: You shouldn't need the `:` after the `-1` here. A more pythonic way for this could be also `dist.endswith('/')` .

joaquimrocha: Also, should we fail when the dist has not slash at the end but the component is still `None`? I am just guessing it is pretty easy to forget the slash at the end, in which case we'd be formatting `packages_url` with a `None` component.

ramcq: Sorted.

joaquimrocha (approved):

PR #10, opened 2018-07-24
Fix parsing more weird manifests

wjt:
I noticed that Endless' internal job that runs this script over a bunch of manifests has been failing for a week or so. In some cases it's because the manifests really do need updating, but in other cases it's because the manifests use features that this script didn't support, namely:

* Referring to a module defined in an external file
* Multi-line strings, which are illegal JSON but accepted by json-glib so naturally are widely-used

### DIFF
--- a/src/checker.py
+++ b/src/checker.py
@@ -28,7 +28,6 @@ class NoManifestCheckersFound(Exception):
     pass
 
 class ManifestChecker:
-
     def __init__(self, manifest):
         self._manifest = manifest
         self._external_data = []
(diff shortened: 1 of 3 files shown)


src/checker.py
```diff
@@ -68,6 +81,11 @@ def _get_finish_args_extra_data_from_json(self, json_data):
     def _get_module_data_from_json(self, json_data):
         external_data = []
         for module in json_data.get('modules', []):
+            if isinstance(module, str):
+                module_path = os.path.join(os.path.dirname(self._manifest),
+                                           module)
```
wjt, line 87: @ramcq for this to work, the `check-flatpak-external-apps` job will have to be modified to recursively clone each repo, rather than just running `wget` on each URL.

PR #11, opened 2018-07-30
Send non-default User-Agent

wjt:
Without this, we see:

   WARNING:root:https://dl.discordapp.net/apps/linux/0.0.5/discord-0.0.5.tar.gz returned HTTP Error 403: Forbidden
   BROKEN: discord.tar.gz
       Unreachable URL: https://dl.discordapp.net/apps/linux/0.0.5/discord-0.0.5.tar.gz

Literally any other User-Agent beside the default seems to work.  This is only strictly needed in check_url_reachable() but it's polite to send a meaningful User-Agent anyway.

I've added a warning when a URL is unreachable including the HTTP error; previously there was no detail in the log output beside "BROKEN" but in fact it was a bug in the script, not an out-of-date URL.


joaquimrocha:
LGTM. Thanks for the fix.

PR #12, opened 2018-09-07
Support YAML manifests

wjt:
Plus some logging stuff so it's clear what's taking so long when you run with `--verbose`, and deleting unused code


joaquimrocha:
Looks great! Thanks a lot!

PR #14, opened 2019-02-07
debianrepochecker: reinitialize apt_pkg each time

wjt:
Without this, the Dir::State::status path is preserved between calls to
this function. Since it's inside a temporary directory that is destroyed
when the function returns, this makes the second and subsequent calls
fail when constructing the apt.Cache():

    apt_pkg.Error: E:Could not open file
    /tmp/tmpvrmj49ts/var/lib/dpkg/status - open (2: No such file or
    directory), E:Problem opening /tmp/tmpvrmj49ts/var/lib/dpkg/status,
    E:The package lists or status file could not be parsed or opened.

This makes the tests pass again (sorry).

### DIFF
--- a/src/checkers/debianrepochecker.py
+++ b/src/checkers/debianrepochecker.py
@@ -118,6 +118,7 @@ class DebianRepoChecker(Checker):
                 pass
 
             # Setup generic configuration
+            apt_pkg.init()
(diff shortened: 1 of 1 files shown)


joaquimrocha (approved):
LGTM.

PR #15, opened 2019-02-26
Support writing back updated JSON manifests

wjt:
Some caveats:

* There's no plumbing to commit the result to Git, push a branch and open a pull request: this still has to be run by hand
* YAML output is not supported, for reasons explained briefly in the relevant commit

But it's a start, and I have used it to update endlessm/eos-google-chrome-app and flathub/com.dropbox.Client.

### DIFF
--- a/src/checker.py
+++ b/src/checker.py
@@ -153,7 +153,7 @@ class ManifestChecker:
         for i, data in enumerate(self._external_data, 1):
             # Ignore if the type is not the one we care about
             if filter_type is not None and filter_type != data.type:
-                continue;
+                continue
 
             log.debug('[%d/%d] checking %s', i, n, data.filename)
 
--- a/src/checker.py
+++ b/src/checker.py
@@ -139,6 +139,8 @@ class ManifestChecker:
 
             for checker in self._checkers:
                 checker.check(data)
+                if data.state != ExternalData.State.UNKNOWN:
+                    break
             ext_data_checked.append(data)
 
         return ext_data_checked
--- a/src/checker.py
+++ b/src/checker.py
@@ -18,8 +18,9 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
 from collections import OrderedDict
+from checkers import ALL_CHECKERS
 from lib.externaldata import (
-    CheckerRegistry, ExternalData, ExternalDataSource, ExternalDataFinishArg,
+    ExternalData, ExternalDataSource, ExternalDataFinishArg,
 )
 
 import json
@@ -43,9 +44,8 @@ class ManifestChecker:
         self._manifest = manifest
         self._external_data = []
 
-        # Load and initialize checkers
-        CheckerRegistry.load(os.path.join(os.path.dirname(__file__), 'checkers'))
-        self._checkers = [checker() for checker in CheckerRegistry.get_checkers()]
+        # Initialize checkers
+        self._checkers = [checker() for checker in ALL_CHECKERS]
 
         # Map from filename to parsed contents of that file. Sources may be
         # specified as references to external files, which is why there can be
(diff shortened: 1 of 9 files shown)


src/checker.py
```diff
@@ -155,3 +164,13 @@ def get_outdated_external_data(self):
             for data in self._external_data
             if data.state == ExternalData.State.BROKEN or data.new_version
         ]
+
+    def update_manifests(self):
+        """Updates references to external data in manifests."""
+        changed = any(data.update() for data in self._external_data)
+        if changed:
```
wjt: This needs to be per-file; otherwise, unchanged files get spuriously reindented. (Found while testing on https://github.com/flathub/com.adobe.Flash-Player-Projector)

src/lib/externaldata.py
```diff
@@ -93,6 +94,64 @@ def to_json(self):
...
         return json.dumps(json_data, indent=4)
 
+
+class ExternalDataSource(ExternalData):
+    def __init__(self, source, data_type, url):
+        name = source.get('filename')
+        if not name:
+            name = source.get('dest-filename')
+        if not name:
+            name = os.path.basename(url)
```
joaquimrocha: You could do all these with `name = result1 or result2 or result3`, right?

src/checker.py
```diff
@@ -86,6 +86,15 @@ def _read_manifest(self, manifest_path):
...
         return contents
 
+    def _dump_manifest(self, manifest_path, contents):
+        _, ext = os.path.splitext(manifest_path)
+        if ext in ('.yaml', '.yml'):
+            raise NotImplementedError(
+                "Updating YAML manifests is not yet supported"
+            )
+        else:
+            return json.dumps(contents, indent=4)
```
joaquimrocha, line 97: AFAIK this may end up giving you slightly a different JSON file than the input (beyond what are the changes). e.g. comments will be stripped of, possibly also the formatting and/or sorting of elements.
These are the main reasons why I didn't implement this solution from the beginning, but even with thos eventual annoyances, it's good to see this feature in now! Kudos!

joaquimrocha (approved):
LGTM! I have left a small nitpick comment, so I am approving the PR but leaving it for your to merge should you want to address that.

src/checker.py
wjt: It actually does preserve element ordering in dicts due to [this](https://github.com/endlessm/flatpak-external-data-checker/blob/master/src/checker.py#L63)¹, but otherwise, yes, it normalises the formatting (which I consider a feature :stuck_out_tongue:) and removes comments (which are useful, sure, but personally I can live without them in exchange for this).

¹ Actually, from Python 3.6, `dict` is ordered, so `json.dumps(json.loads(s))` never reorders keys even if you don't specify `object_pairs_hook`.

joaquimrocha: Awesome! Thanks for clarifying. I didn't see an answer to my other comment, so I am keeping this unmerged and let you decide.

src/lib/externaldata.py
wjt: Yep, this looks nicer; good point.

wjt:
Pushed a combination of new patches and new versions of old patches. I assumed I'd get a handy "force-pushed" link but no... here's the diff https://github.com/endlessm/flatpak-external-data-checker/compare/3ec283f72344deff8952d81795410b74cbcfa6c5..T19948-update-JSON-manifest.

wjt:
Two more patches to support checking https://github.com/flathub/com.adobe.Flash-Player-Projector

wjt:
I dropped the scrapechecker one. It seems like one should try to use a service like https://release-monitoring.org/project/17854/ rather than inventing something new, but this is classic scope creep…!

wjt:
Timing out and merging this. I have used it many times to update Chrome and Dropbox, and c4953e9902bafa7a9067248fa8a67024101a55c4 would have alerted us that https://github.com/flathub/com.viber.Viber/pull/11 was necessary.

LeandroStanger:
https://github.com/flathub/com.wps.Office/pull/6

PR #19, opened 2019-06-06
Commit results and open PRs when run with --update

wjt:
This needs:

* an SSH key that can push to GitHub
* the $GITHUB_TOKEN environment variable set

Supersedes #18.


wjt:
I use this by hand and it seems to work.

PR #20, opened 2019-06-18
CI+docker

andrunko:
Add support to run the checker in a container and define jenkins job using a Jenkinsfile. 

Job at https://ci.endlessm-sf.com/job/check-flatpak-external-apps-pipeline/.

This does not yet passes '--update' to the checker as I'm unsure on how to get the `GITHUB_TOKEN` required to use this option.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,48 @@
+node('flatpak-builder') {
+    def image
+
+    // Global env variables used by git
+    env.GIT_AUTHOR_NAME='Endless External Data Checker'
+    env.GIT_COMMITTER_NAME='Endless External Data Checker'
+    env.GIT_AUTHOR_EMAIL='desktop@endlessm.com'
+    env.GIT_COMMITTER_EMAIL='desktop@endlessm.com'
+
+    stage('Clone repository') {
+        checkout scm
+    }
+
+    stage('Build container') {
+        /* FIXME: We pass the user and group ids when building the image as
+         * a workaround to make sure the user running the container exists
+         * inside of it (see Dockerfile). This is needed because jenkins runs
+         * the container with the same user as the host (with '-u <uid>:<gid>')
+         * but 'git push' fails with 'No user exists for uid ...' if the user
+         * doesn't exist inside the container (other option is to mount
+         * '/etc/passwd' as a volume and export HOME/USER envs accordingly
+         * when running the container ('image.inside' below)).
+         *
+         * See also https://issues.jenkins-ci.org/browse/JENKINS-47026.
+         */
+        image = docker.build('flatpak-external-data-checker:latest',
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,50 @@
...
+    env.GIT_COMMITTER_EMAIL='emdev@endlessm.com'
+
+    stage('Clone repository') {
+        checkout scm
+    }
+
+    stage('Build container') {
+        /* FIXME: We pass the user and group ids when building the image as
+         * a workaround to make sure the user running the container exists
+         * inside of it (see Dockerfile). This is needed cause jenkins runs
```
wjt: s/cause/because/ FYI

Jenkinsfile
```diff
@@ -0,0 +1,50 @@
...
+         *
+         * See also https://issues.jenkins-ci.org/browse/JENKINS-47026.
+         */
+        image = docker.build('flatpak-external-data-checker:latest',
+                             '--build-arg USER_ID=$(id -u) --build-arg GROUP_ID=$(id -g) .')
+    }
+
+    stage('Run checker in container') {
+        try {
+            /* The '--privileged' param is required to run 'bwrap' which is used
```
wjt, line 32: We could make `bwrap` optional, I guess. OTOH it is probably still worth confining the arbitrary code execution in question to not have access to our Git credentials – does it actually need to be tweaked further to not inherit SSH credentials etc. from its environment?

wjt (commented):
This looks great, thanks for looking into this. Just one question around the use of `--privileged`

Jenkinsfile
andrunko: The `sshagent` instruction makes the checker inherit our SSH credentials (I added it given when using `--update` it will need access to push to git iiuc). We could remove the `sshagent` instruction (or change it to a credential that only allows pulling if needed) if we dont need support for publishing to git.

Jenkinsfile
andrunko: Thanks, fixed.

Jenkinsfile
wjt: I understand that this is how the checker itself inherits these credentials, and yes, we want this. What I'm wondering is: is https://github.com/endlessm/flatpak-external-data-checker/blob/master/src/lib/utils.py#L71-L81 enough to confine the arbitrary 3rd-party code that function is executing from being able to do anything nefarious with those creds, or do we also need some `--unset-env` and/or masking off of the ssh-agent?

andrunko: hmm, I see - tbh I haven't checked this but to do that we probably want to unset SSH_AUTH_SOCK here - needs to test to see if it works though.

andrunko:
Force pushed to update the committer email to desktop@.

wjt (approved):

PR #25, opened 2019-07-10
Add Flash extra data checker

refi64:
I, uhh, pushed this yesterday and supposedly submitted a PR through [hub](https://hub.github.com/), which apparently didn't work, that's what I get for trusted technology I guess?

### DIFF
--- a/src/checkers/flashchecker.py
+++ b/src/checkers/flashchecker.py
@@ -1,0 +1,119 @@
+# Flash Checker: A checker that to see if the url is pointing to the latest Flash Player.
+#
+# Consult the README for information on how to use this checker.
+#
+# Copyright © 2019 Endless Mobile, Inc.
+#
+# Authors:
+#       Ryan Gonzalez <rymg19@gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+import json
+import logging
+import urllib.error
+import urllib.request
+import urllib.parse
+
+from lib.externaldata import ExternalData, ExternalFile, Checker
+from lib import utils
+
+log = logging.getLogger(__name__)
+
+
+FLASH_BASE_URL = 'http://get.adobe.com/flashplayer/webservices/json/?{params}'
+
+FLATPAK_TO_FLASH_ARCH_MAP = {
+    'i386': 'x86-32',
+    'x86_64': 'x86-64',
+}
+
+BROWSER_TO_PAPI_MAP = {
+    'Chrome': 'ppapi',
+    'Firefox': 'npapi',
+}
+
+
+class FlashChecker(Checker):
+    def _should_check(self, external_data):
+        return external_data.checker_data.get('type') == 'flash'
+
+    def _get_arches(self, external_data):
+        if len(external_data.arches) != 1:
+            return None
+
+        flatpak_arch = external_data.arches[0]
+        return flatpak_arch, FLATPAK_TO_FLASH_ARCH_MAP.get(external_data.arches[0])
+
+    def check(self, external_data):
+        if not self._should_check(external_data):
+            log.debug('%s is not a flash type ext data', external_data.filename)
+            return
+
+        browser = external_data.checker_data['browser'].title()
+
+        try:
+            papi = BROWSER_TO_PAPI_MAP[browser]
+        except KeyError:
+            log.warning('%s has an invalid browser (should be one of %s)',
+                        external_data.filename, ', '.join(BROWSER_TO_PAPI_MAP))
+            return
+
+        arches = self._get_arches(external_data)
+        if arches is None:
+            log.warning('%s has invalid only-arches (should be one of %s)',
+                        external_data.filename,
+                        ', '.join(f'[{arch}]' for arch in FLATPAK_TO_FLASH_ARCH_MAP))
+            return
+
+        flatpak_arch, flash_arch = arches
+
+        request_params = {'platform_type': 'Linux', 'eventname': 'flashplayerotherversions',
+                          'platform_arch': flash_arch}
+        request_url = FLASH_BASE_URL.format(params=urllib.parse.urlencode(request_params))
+        with urllib.request.urlopen(request_url) as resp:
+            latest_version_data = json.load(resp)
+
+        latest_version = None
+        latest_url = None
+
+        for version in latest_version_data:
+            if (version['installation_type'] == 'Standalone' and version['browser'] == browser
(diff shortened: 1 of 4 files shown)


src/checkers/flashchecker.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+FLASH_BASE_URL = 'http://get.adobe.com/flashplayer/webservices/json/?{params}'
+
+FLATPAK_TO_FLASH_ARCH_MAP = {
+    'i386': 'x86-32',
+    'x86_64': 'x86-64',
+}
+
+BROWSER_TO_PAPI_MAP = {
```
wjt, line 43: Why not just specify `ppapi`/`npapi` in the checker data?

src/checkers/flashchecker.py
```diff
@@ -0,0 +1,130 @@
...
+        try:
+            papi = BROWSER_TO_PAPI_MAP[browser]
+        except KeyError:
+            log.warning('%s has an invalid browser (should be Chrome or Firefox)',
+                        external_data.filename)
+            return
+
+        arches = self._get_arches(external_data)
+        if arches is None:
+            log.warning('%s has invalid only-arches (should be [i386] or [x86_64])',
```
wjt: Derive this error message from the keys of `FLATPAK_TO_FLASH_ARCH_MAP`

src/checkers/flashchecker.py
```diff
@@ -0,0 +1,130 @@
...
+        except Exception:
+            log.exception('Unexpected exception while checking %s', url)
+            external_data.state = ExternalData.State.BROKEN
+        else:
+            new_version = ExternalFile(latest_url, checksum, size, latest_version)
+
+            if external_data.current_version.matches(new_version):
+                external_data.state = ExternalData.State.VALID
+            else:
+                external_data.state = ExternalData.State.BROKEN
```
wjt: I don't think it's quite right to set it as broken here – the old URL may still work. It's somewhat academic though

src/checkers/flashchecker.py
```diff
@@ -0,0 +1,130 @@
+# Flash Checker: A checker that to see if the url is pointing to the latest Flash Player.
+#
+# The contens of the x-checker-data for the module should be .e.g:
```
wjt:
```suggestion
# The x-checker-data for the module should be, for example:
```

wjt: I'd rather this documentation lived in the `README` instead/as well.

src/checkers/flashchecker.py
```diff
@@ -0,0 +1,130 @@
...
+            if (version['installation_type'] == 'Standalone' and version['browser'] == browser
+                    and version['installer_architecture'] == flash_arch):
+                assert version['platform'] == 'Linux'
+
+                latest_version = version['Version']
+                latest_url = version['download_url']
+                break
+
+        if latest_version is None:
+            log.warning(f'%s had no available URL', external_data.filename)
```
wjt:
```suggestion
            log.warning('%s had no available URL', external_data.filename)
```

No need for an f-string here

wjt (changes requested):
I just had a cursory look over this

src/checkers/flashchecker.py
```diff
@@ -0,0 +1,130 @@
...
+
+BROWSER_TO_PAPI_MAP = {
+    'Chrome': 'ppapi',
+    'Firefox': 'npapi',
+}
+
+
+class FlashChecker(Checker):
+    def __init__(self):
+        self._pkgs_cache = {}
```
wjt: This is unused.

src/checkers/flashchecker.py
refi64: Later in the code we have need the browser name anyway to find the proper URL, and having to specify both the browser *and* the plugin API feels redundant when one is always associated with the other.

I mean it could be the other way around (specify the plugin API and derive the browser name), but using the browser name felt clearer.

src/checkers/flashchecker.py
refi64: Ah, I didn't realize it'll still show a new version if it's not marked as broken.

src/checkers/flashchecker.py
refi64: Whoops, force of habit...

src/checkers/flashchecker.py
refi64: Err I actually copied that format from debianrepochecker.py without noticing the typo, should I fix it there as well?

wjt: Yes please

src/checkers/flashchecker.py
wjt: (I think it will but I could be wrong!)

src/checkers/flashchecker.py
```diff
@@ -0,0 +1,118 @@
...
+        if not self._should_check(external_data):
+            log.debug('%s is not a flash type ext data', external_data.filename)
+            return
+
+        browser = external_data.checker_data['browser'].title()
+
+        try:
+            papi = BROWSER_TO_PAPI_MAP[browser]
+        except KeyError:
+            log.warning('%s has an invalid browser (should be Chrome or Firefox)',
```
wjt: Derive this from the keys in that map

src/checkers/flashchecker.py
```diff
@@ -0,0 +1,118 @@
...
+                        external_data.filename)
+            return
+
+        arches = self._get_arches(external_data)
+        if arches is None:
+            log.warning('%s has invalid only-arches (should be one of %s)',
+                        external_data.filename,
+                        ', '.join(f'[{arch}]' for arch in FLATPAK_TO_FLASH_ARCH_MAP))
```
wjt, line 78: Why wrap these in '[]'?

src/checkers/flashchecker.py
```diff
@@ -0,0 +1,118 @@
...
+        for version in latest_version_data:
+            if (version['installation_type'] == 'Standalone' and version['browser'] == browser
+                    and version['installer_architecture'] == flash_arch):
+                assert version['platform'] == 'Linux'
+
+                latest_version = version['Version']
+                latest_url = version['download_url']
+                break
+
+        if latest_version is None:
```
wjt: I am keen on spelling this kind of loop as:

```python3
for version in latest_version_data:
    if ...:
         break
else:
    # this block runs if the break is never reached
```

however I know this is not a universally-popular Python feature :-)

src/checkers/flashchecker.py
```diff
@@ -0,0 +1,118 @@
...
+            log.warning('%s returned %s', url, e)
+            external_data.state = ExternalData.State.BROKEN
+        except Exception:
+            log.exception('Unexpected exception while checking %s', url)
+            external_data.state = ExternalData.State.BROKEN
+        else:
+            external_data.state = ExternalData.State.VALID
+            new_version = ExternalFile(latest_url, checksum, size, latest_version)
+            if not external_data.current_version.matches(new_version):
+                external_data.new_version = new_version
```
wjt, line 119: Don't we want to set this in the two exception paths above, too? If the current URL is unreachable, *and* there's a new version, we want to update to the new version.

refi64: This path is taken if getting the information of the *new* URL fails, so if either of the exception blocks are taken, we won't have enough information to set the new version with.

src/checkers/flashchecker.py
refi64: Since `only-arches` has to be a list, this is supposed to show the YAML-like syntax for having a one-item list of the desired values. If you want, I can change it to add double-quotes, like JSON syntax.

src/checkers/flashchecker.py
refi64: Ah too much C++ lately, forgot all about that.

wjt (approved):

Issue #17, opened 2019-05-20
Support updating <releases> in appdata


wjt:
Flathub now strongly encourages including release information in appdata; it is a hard requirement for PRs, and someday will be required for real builds, too. So this tool will have to learn to update that in cases where the appdata is not provided by upstream.

There are a number of ways we might determine the version number:

* Some packages include the version number in their URL (or are Debian packages with a structured version number)
* Some packages might have entries on release-monitoring.org which includes version information #16
* AppImages can be unpacked by passing `--appimage-extract`; the `.desktop` file within includes an `X-AppImage-Version` key

wjt:
I implemented the `--appimage-extract` stuff on #18

muelli:
for actually updating the XML file in place, something like

```
xmlstarlet ed --ps --inplace -L -s /component/releases -t elem -n releaseTMP -v "" \
    -i //releaseTMP -t attr -n "version" -v "20190801" \
    -i //releaseTMP -t attr -n "date" -v "2019-08-01" \
    -a //releaseTMP -t text -n text -v "description" \
    -r //releaseTMP -v release \
    *appdata.xml
```
might be handy.

wjt:
I ended up writing my own bespoke in-place XML editor using Python's `xml.sax` API because I couldn't see a way to match indentation with `xmlstarlet`, and `xml.dom.minidom` can't preserve attribute ordering. Not my favourite activity but it passed the time on a delayed flight.

muelli:
I was wondering whether we'd want a separate tool for adding a release to an appdata file.
Something like `appdata-add-release --description-from-file ./description   version  date` or so.
what do you think about that?

wjt:
Such a tool sounds potentially useful for app maintainers, and I don't see why it couldn't live in this repo. I'm keen for the checker itself to be as non-interactive as possible – I think it is likely to be difficult to find a sufficiently-generic way to automatically add a description!

PR #26, opened 2019-08-01
Support scraping version numbers from URLs, plus miscellaneous fixes

wjt:
To make a start on #17 for Dropbox, I added a way to scrape version numbers out of URLs with a regex. Then while doing a bit of supporting refactoring, I found the flashchecker has no tests, added one, and found a couple of bugs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,9 +127,20 @@ the latest version:
 ```json
   "x-checker-data": {
                      "type": "rotating-url",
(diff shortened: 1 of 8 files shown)


wjt:
@refi64 I can't tag you for review but you may be interested :)

andrunko (approved):
lgtm

PR #27, opened 2019-08-01
Update manifests for firefox/flash extension on external data checker

andrunko:
https://phabricator.endlessm.com/T22543

### DIFF
--- a/src/checker.py
+++ b/src/checker.py
@@ -125,8 +125,17 @@ class ManifestChecker:
                 module_path = path
 
             sources = module.get('sources', [])
+            inline_sources = [ source for source in sources if not isinstance(source, str) ]
+            external_sources = [ source for source in sources if isinstance(source, str) ]
+
             external_data = self._external_data.setdefault(module_path, [])
-            external_data.extend(ExternalDataSource.from_sources(sources))
+            external_data.extend(ExternalDataSource.from_sources(inline_sources))
+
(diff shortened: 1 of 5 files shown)


src/checker.py
```diff
@@ -125,8 +125,18 @@ def _get_module_data_from_json(self, path, json_data):
...
             external_data = self._external_data.setdefault(module_path, [])
-            external_data.extend(ExternalDataSource.from_sources(sources))
+            external_data.extend(ExternalDataSource.from_sources(inline_sources))
+
+            for external_source in external_sources:
+                external_source_path = os.path.join(os.path.dirname(self._manifest),
+                                                    external_source)
+                external_source_data = self._read_manifest(external_source_path)
+                external_data = self._external_data.setdefault(external_source_path, [])
+                external_data.extend(ExternalDataSource.from_sources(external_source_data))
```
wjt: I think you can just write:

```python3
self._external_data[external_source_path] = ExternalDataSource.from_sources(external_source_data)
```

The `setdefault` above is needed because the body of this loop runs multiple times for the same file. I think that if the same external file is referenced more than once, this would still do the right thing, because we only need to check & update the file once.

wjt (changes requested):
One minor thing.

src/checker.py
andrunko: Done.

PR #28, opened 2019-08-07
Add specialized checker for firefox + some small improvements

andrunko:
https://phabricator.endlessm.com/T21208

### DIFF
--- a/src/checker.py
+++ b/src/checker.py
@@ -113,7 +113,7 @@ class ManifestChecker:
     def _get_finish_args_extra_data_from_json(self, path, json_data):
         finish_args = json_data.get('finish-args', [])
         external_data = self._external_data.setdefault(path, [])
-        external_data.extend(ExternalDataFinishArg.from_args(finish_args))
+        external_data.extend(ExternalDataFinishArg.from_args(path, finish_args))
 
     def _get_module_data_from_json(self, path, json_data):
         for module in json_data.get('modules', []):
@@ -129,13 +129,13 @@ class ManifestChecker:
             external_sources = [ source for source in sources if isinstance(source, str) ]
 
             external_data = self._external_data.setdefault(module_path, [])
-            external_data.extend(ExternalDataSource.from_sources(inline_sources))
+            external_data.extend(ExternalDataSource.from_sources(module_path, inline_sources))
 
             for external_source in external_sources:
                 external_source_path = os.path.join(os.path.dirname(self._manifest),
                                                     external_source)
                 external_source_data = self._read_manifest(external_source_path)
-                self._external_data[external_source_path] = ExternalDataSource.from_sources(external_source_data)
+                self._external_data[external_source_path] = ExternalDataSource.from_sources(external_source_path, external_source_data)
 
     def check(self, filter_type=None):
         '''Perform the check for all the external data in the manifest
--- a/src/checker.py
+++ b/src/checker.py
@@ -21,7 +21,7 @@
 from collections import OrderedDict
 from checkers import ALL_CHECKERS
 from lib.externaldata import (
-    ExternalData, ExternalDataSource, ExternalDataFinishArg,
+    ModuleData, ExternalData, ExternalDataSource, ExternalDataFinishArg,
 )
 
 import json
@@ -49,6 +49,7 @@ class ManifestChecker:
 
     def __init__(self, manifest):
         self._manifest = manifest
+        self._modules_data = {}
         self._external_data = {}
 
         # Initialize checkers
@@ -124,18 +125,27 @@ class ManifestChecker:
             else:
                 module_path = path
 
+            module_name = module.get('name')
+            module_data = ModuleData(module_name, module_path, module)
+
             sources = module.get('sources', [])
             inline_sources = [ source for source in sources if not isinstance(source, str) ]
             external_sources = [ source for source in sources if isinstance(source, str) ]
 
             external_data = self._external_data.setdefault(module_path, [])
-            external_data.extend(ExternalDataSource.from_sources(module_path, inline_sources))
+            datas = ExternalDataSource.from_sources(module_path, inline_sources)
+            external_data.extend(datas)
+            module_data.external_data.extend(datas)
 
             for external_source in external_sources:
                 external_source_path = os.path.join(os.path.dirname(self._manifest),
                                                     external_source)
                 external_source_data = self._read_manifest(external_source_path)
-                self._external_data[external_source_path] = ExternalDataSource.from_sources(external_source_path, external_source_data)
+                datas = ExternalDataSource.from_sources(external_source_path, external_source_data)
+                self._external_data[external_source_path] = datas
+                module_data.external_data.extend(datas)
+
+            self._modules_data[module_name] = module_data
 
     def check(self, filter_type=None):
         '''Perform the check for all the external data in the manifest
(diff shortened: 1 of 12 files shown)


andrunko:
Force pushed to add `--rm` to `podman run` in the `run-in-container.sh` script.

src/flatpak-external-data-checker
```diff
@@ -183,6 +183,10 @@ def main():
                         help="Update manifest(s) to refer to new versions of "
                              "external data",
                         action="store_true")
+    parser.add_argument("--commit-only",
+                        help="Do not open PRs for updates, only commit changes "
+                             "to external data (requires --update)",
```
wjt: I think it should imply `--update`, rather than documenting that it does nothing unless you pass another flag.

src/lib/externaldata.py
```diff
@@ -74,9 +74,9 @@ def __str__(self):
                '  Checker: {checker_data}'.format(state=self.state.name,
                                                   filename=self.filename,
                                                   type=self.type.name,
-                                                  url=self.url,
-                                                  checksum=self.checksum,
-                                                  size=self.size,
+                                                  url=self.new_version.url if self.new_version else self.current_version.url,
```
wjt: Personally I would prefer (eg):

```python3
version = self.new_version or self.current_version
...
    url=version.url,
    checksum=version.checksum,
...
```

but, not a big deal.

src/lib/externaldata.py
```diff
@@ -137,36 +137,41 @@ def update(self):
...
-            for i, arg in enumerate(finish_args)
+            cls(source_path, arg, finish_args)
+            for arg in finish_args
             if arg.startswith(cls.PREFIX)
         ]
 
     def update(self):
         if self.new_version is not None:
+            if not self.source in self.source_parent:
+                log.warning("Unable to find index for %s", self.source)
```
wjt: This should only happen on programmer error – I'd just leave this check out and allow `index()` to throw an exception.

tests/test_checker.py
```diff
@@ -69,7 +69,9 @@ def test_check_filtered(self):
 
         ext_data = dummy_checker.check()
         ext_data_from_getter = dummy_checker.get_external_data()
-        self.assertEqual(ext_data_from_getter, ext_data)
+        self.assertEqual(len(ext_data), len(ext_data_from_getter))
+        for data in ext_data:
+            self.assertIn(data, ext_data_from_getter)
```
wjt: Why is this change needed – is the order no longer preserved?

src/checkers/firefoxchecker.py
```diff
@@ -0,0 +1,181 @@
...
+        for line in sha256_table.splitlines():
+            sha256, path = line.split(None, 1)
+            path = path.strip()
+
+            if path.startswith(self.FIREFOX_PLATFORM):
+                assert path.count('/') == 2, path
+                _, dirname, filename = path.split('/')
+
+                source_filename = None
+                if re.match(checker_data['browser']['filename-match'], filename) and \
```
wjt: Why do we need to specify regular expressions to match in the manifest? This checker is Firefox-specific. We can just hardcode them.

src/checkers/firefoxchecker.py
```diff
@@ -0,0 +1,181 @@
...
+                }
+                # add new data to the same manifest as the main browser tarball
+                data = ExternalDataSource.from_source(browser_data.source_path, source, browser_data.source_parent)
+                data.state = ExternalData.State.ADDED
+                added.append(data)
+
+        return added
+
+    def _get_latest_available_version(self):
+        if self._latest_version:
```
wjt: I think you can drop all the manual caching and decorate this function with [functools.lru_cache()](https://docs.python.org/3/library/functools.html#functools.lru_cache):

```python3
    @functools.lru_cache()
    def _get_latest_available_version(self):
```

src/checkers/firefoxchecker.py
```diff
@@ -0,0 +1,181 @@
...
+
+        latest_version_data = {}
+        request_url = self.FIREFOX_RELEASE_INFO_URL
+        with urllib.request.urlopen(request_url) as resp:
+            latest_version_data = json.load(resp)
+
+        if isinstance(latest_version_data, dict) and 'LATEST_FIREFOX_VERSION' in latest_version_data:
+            self._latest_version = latest_version_data['LATEST_FIREFOX_VERSION']
+        else:
+            self._latest_version = None
```
wjt: We should fail in this case.

wjt (changes requested):

tests/test_checker.py
andrunko: Yes, the order is not preserved on the return value of `ManifestChecker.check()` anymore. That is because we run the modules check first and add all modules external data to the check list and then we check individual sources that were not processed yet and add them as they get processed (opposed to only checking sources in order as it was before).

As far as I could see nothing (apart from this test) was using the return value of `ManifestChecker.check()`, so I think it should be safe but please let me know.

src/checkers/firefoxchecker.py
andrunko: This method is invoked inside a `try/catch/finally` and the `self._latest_version` is checked on `finally` with a warning if set to `None`.

Do you think we should raise an exception (i.e. break the checker) in this case? Or how else 
 would you fail here?

src/checkers/firefoxchecker.py
andrunko: Ok, tbh I prefer to have it configurable such that we don't have to force sources/translations filename to match (e.g. in a future official firefox flatpak). 

Maybe leave the checker options but use default values if they are not set?

src/lib/externaldata.py
andrunko: Ack.

src/flatpak-external-data-checker
andrunko: Ack.

src/checkers/firefoxchecker.py
andrunko: Ok, but isn't this more expensive than the current cache? Not that we care much about resources here but still...

src/checkers/firefoxchecker.py
wjt: > Do you think we should raise an exception (i.e. break the checker) in this case?

Probably! If this web service turns out to be unreliable we can do something different.

src/checkers/firefoxchecker.py
wjt: It'll be a dictionary whose key is `self` and whose value is the return value – a dozen bytes or so. I think this is worth the code size reduction.

src/checkers/firefoxchecker.py
wjt: A future official Firefox Flatpak will be generated as part of the Mozilla release process so will not need this script.

wjt: > Maybe leave the checker options but use default values if they are not set?

OK! I predict that we'll never need to set them but I am prepared to be wrong :-)

tests/test_checker.py
wjt: It's probably fine but perhaps you could express this as:

```
         self.assertEqual(len(ext_data), len(ext_data_from_getter))
         self.assertEqual(set(ext_data), set(ext_data_from_getter))
```

?

src/flatpak-external-data-checker
andrunko: Done.

src/lib/externaldata.py
andrunko: Done.

src/lib/externaldata.py
andrunko: Done.

tests/test_checker.py
andrunko: Done.

src/checkers/firefoxchecker.py
andrunko: Done, ended up hardcoding the expressions in the checker - if in the future we find the need we can make it configurable again.

src/checkers/firefoxchecker.py
andrunko: Done.

src/checkers/firefoxchecker.py
andrunko: Done.

andrunko:
Everything should be updated, please check again.

andrunko:
@wjt Squashed the fixup for the script permissions, thanks!

src/checkers/firefoxchecker.py
```diff
@@ -0,0 +1,162 @@
...
+
+        log.debug('Retrieving latest firefox version info')
+        latest_info = None
+        try:
+            latest_info = self._get_latest_version_info(latest_version)
+        except Exception as e:
+            log.warning('Error retrieving latest firefox version info: %s', e)
+        finally:
+            if not latest_info:
+                log.warning('Could not retrieve latest firefox version information')
```
wjt: Again I would prefer to skip the exception handling and just let it bubble up in this case. We won't notice the warnings, but we will notice the job failing.

src/checker.py
```diff
@@ -189,15 +227,24 @@ def get_outdated_external_data(self):
...
         for data in datas:
-            if data.new_version is None:
+            if data.new_version is None and \
+               data.state != ExternalData.State.ADDED and \
+               data.state != ExternalData.State.REMOVED:
                 continue
 
             data.update()
+            if data.state == ExternalData.State.ADDED:
+                message = "Added {}".format(data.filename)
```
wjt, line 245: I guess we'll find out whether, in practice, new translations are added/removed within the lifetime of a given Firefox release. I would expect not.

It would be nice if the commit message could read “Update Firefox to X.Y.Z” rather than “Update 94 modules”; and this would also make #17 easier because we would have already boiled the update down to the salient fact. But, as Voltaire probably didn't say, RESOLVED LATER is the friend of good!

Each time Firefox is updated, the commit message will be 180 lines long because every file will be both added (new version) and removed (old version) whereas the logical change is "Update Firefox to ____". I feel like it would be nice to be able to do

wjt (approved):
I had a few other comments but I just fixed them directly.

src/checker.py
andrunko: Makes sense, I will have a think about it - we could probably use the module the updated data comes from to build the commit msg.

PR #34, opened 2019-08-27
Update <releases> in appdata

### LINKED ISSUE #17: Support updating <releases> in appdata

wjt:
Doing this without generating a huge diff was a lot more involved than I
expected, but this seems to work OK now.

I have tested this locally against:

* endlessm/eos-google-chrome-app, which currently has no `<releases>`: the checker adds it
* flathub/us.zoom.Zoom, which already has `<releases>`: it prepends a new one, albeit with the attributes in a different order to the existing entries. There's a test marked `expectedFailure` for this; the order has no semantic impact and the 18 repos I checked have no trend towards one or another.

Fixes #17.

### DIFF
--- a/src/checker.py
+++ b/src/checker.py
@@ -20,10 +20,12 @@
 
 from collections import OrderedDict
 from checkers import ALL_CHECKERS
+from lib.appdata import add_release_to_file
 from lib.externaldata import (
     ModuleData, ExternalData, ExternalDataSource, ExternalDataFinishArg,
 )
 
+import datetime as dt
 import json
 import logging
 import os
@@ -234,6 +236,8 @@ class ManifestChecker:
         ]
 
     def _update_manifest(self, path, datas, changes):
+        last_update = None
+
         for data in datas:
             if data.new_version is None and \
                data.state != ExternalData.State.ADDED and \
@@ -249,14 +253,23 @@ class ManifestChecker:
                 message = "Update {} to {}".format(
                     data.filename, data.new_version.version
                 )
+                last_update = data.new_version
             else:
                 message = "Update {}".format(data.filename)
 
             changes[message] = None
 
         if changes:
-            print("Updating {}".format(path))
+            log.info("Updating %s", path)
             self._dump_manifest(path)
+
+            appdata = os.path.splitext(path)[0] + ".appdata.xml"
+            if last_update is not None and os.path.exists(appdata):
+                # TODO: this assumes that the last changed source for which we can
+                # detect a version number is the one corresponding to the application
+                # as a whole. In practice, this is currently true, but in general it
+                # may not be.
+                add_release_to_file(appdata, last_update.version, last_update.timestamp.strftime("%F"))
 
     def update_manifests(self):
         """Updates references to external data in manifests."""
--- a/src/lib/appdata.py
+++ b/src/lib/appdata.py
@@ -1,0 +1,145 @@
+# Copyright © 2019 Endless Mobile, Inc.
+#
+# Authors:
+#       Will Thompson <wjt@endlessm.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+"""
+Add a new <release> at the start of the <releases> element in an appdata file,
+preserving as much formatting as is feasible and inserting that element if it is
+missing.
+"""
+
+from io import StringIO
+from xml.sax import make_parser
+from xml.sax.handler import property_lexical_handler
+from xml.sax.saxutils import XMLFilterBase, XMLGenerator
+
+
+class AddVersionFilter(XMLFilterBase):
+    def __init__(self, version, date, parent=None):
+        super().__init__(parent)
+
+        self._version = version
+        self._date = date
+        self._context = []
+        self._emitted_release = False
+        self._releases_padding = ""
+
+    @property
+    def outside_root_element(self):
+        return not self._context
+
+    @property
+    def _in_releases(self):
+        return self._context[1:] == ["releases"]
+
+    def _emit_release(self):
+        super().startElement("release", {"version": self._version, "date": self._date})
+        # TODO: add placeholder <description> if other <release> entries have one
+        super().endElement("release")
+        super().characters(self._releases_padding)
+        self._emitted_release = True
+
+    def startElement(self, name, attrs):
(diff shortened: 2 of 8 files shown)


src/checker.py
```diff
@@ -249,15 +253,24 @@ def _update_manifest(self, path, datas, changes):
...
-            print("Updating {}".format(path))
+            log.info("Updating %s", path)
             self._dump_manifest(path)
 
+            appdata = os.path.splitext(path)[0] + ".appdata.xml"
+            if last_version is not None and os.path.exists(appdata):
+                # This assumes that the last changed source for which we can detect a
+                # new version is the interesting one – this may or may not be true!
+                # TODO: extract date in checkers
+                release_date = dt.date.today()
```
andrunko: hmm, ideally we should have a way to get the latest release info (version+date) for a "top" manifest (from path) instead, but while we don't support that I think using the last changed source is Ok, although I can imagine a situation where we have multiple external data/sources using different versions.

On possible option would be to grow support for specialized manifest checkers (alongside extra-data checkers), with each manifest checker running the external data checkers for that component + providing release info.

src/lib/appdata.py
```diff
@@ -0,0 +1,169 @@
...
+        releases.insertBefore(whitespace.cloneNode(False), new_release)
+
+    out.write(releases.toxml("UTF-8").decode("UTF-8"))
+
+This is a nice concise function, and it does preserve almost all formatting.
+But, attribute order is not preserved, because ... minidom.Element.writexml()
+unconditionally sorts the attributes by name! Thanks!
+
+Instead, I am disappointed to present the following SAX monstrosity.
+"""
```
andrunko, line 24: nit: missing newline

src/lib/appdata.py
```diff
@@ -0,0 +1,169 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+"""
```
andrunko, line 20: nit: missing newline

src/lib/appdata.py
```diff
@@ -0,0 +1,169 @@
...
+    new_release = document.createElement("release")
+    new_release.setAttribute("version", version)
+    new_release.setAttribute("date", date)
+
+    releases.insertBefore(new_release, releases.firstChild)
+    whitespace = previous_release.previousSibling
+    if whitespace is not None and whitespace.nodeType == minidom.Node.TEXT_NODE:
+        releases.insertBefore(whitespace.cloneNode(False), new_release)
+
+    out.write(releases.toxml("UTF-8").decode("UTF-8"))
```
andrunko: Not sure this snippet adds anything here, maybe something for a commit msg instead?

src/lib/appdata.py
```diff
@@ -0,0 +1,169 @@
...
+def add_release(in_, out, version, date):
+    reader = AddVersionFilter(version, date, make_parser())
+    lexical_handler = VerbatimLexicalHandler(reader, out)
+    handler = XMLGenerator(out, encoding="UTF-8", short_empty_elements=True)
+    reader.setContentHandler(handler)
+    reader.setProperty(property_lexical_handler, lexical_handler)
+    reader.parse(in_)
+
+
+def add_release_to_file(path, version, data):
```
andrunko: s/data/date

src/lib/appdata.py
```diff
@@ -0,0 +1,169 @@
...
+        raise NotImplemented
+
+    def endCDATA(self, *args):
+        raise NotImplemented
+
+    def endDTD(self, *args):
+        raise NotImplemented
+
+
+def add_release(in_, out, version, date):
```
andrunko: I would probably s/in_/path/ here as well, or even `appdata_path`.

tests/test_appdata.py
```diff
@@ -0,0 +1,273 @@
...
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
```
andrunko, line 19: nit: missing newline

andrunko (changes requested):

src/checker.py
wjt: Agreed. I realised that `date.today()` is actually not good enough: the branch name used depends on the contents of the tree, so it would change every day and file a duplicate PR until one is merged. Unfortunately the Debian repo format doesn't include timestamps, so I make a `HEAD` request against the package URL and use its timestamp.

src/lib/appdata.py
wjt: Moved to commit messaeg.

src/lib/appdata.py
wjt: Actually it doesn't have to be a path. It can be a file-like object too, and the test suite relies on this. I've made this explicit in the name.

src/checker.py
andrunko: Another option would be to extract the date from the package changelog itself (which may require fetching/extracting the package if python apt doesn't provide that) and for firefox we could get the date from `LAST_RELEASE_DATE` returned on https://product-details.mozilla.org/1.0/firefox_versions.json.

andrunko (approved):
lgtm with a small comment regarding the release date/timestamps for debian packages and firefox. I will leave it to you to decide whether you want to work on it (I am fine to do it as a follow-up change).

src/checker.py
wjt: `python-apt` has some support for fetching changelogs for Debian and Ubuntu, but we don't have a changelog server. https://salsa.debian.org/apt-team/python-apt/blob/master/apt/package.py#L1245-1417

The other option would be to fetch the package (or just the relevant half of the source package – if we have a source package, that is!) and parse the changelog (if present). Personally, I think we can make do with the `HEAD` request for now. I'll add a TODO about this.

---

https://product-details.mozilla.org/1.0/firefox_versions.json has `"LAST_RELEASE_DATE": "2019-07-09"` but https://www.mozilla.org/en-US/firefox/68.0.2/releasenotes/ says:

> Version 68.0.2, first offered to Release channel users on August 14, 2019

So I don't think that's going to fly. (I'll add a note about this.) However, we can and should propagate the version number from that file into the `ExternalFile` object. I'll make that change.

While looking at this, I realised that we currently fetch all the files for Firefox, but we don't need to – we get their paths from `SHA256SUMS` so we have those, and we can just make a `HEAD` request to get the size. That's another task though.

PR #37, opened 2019-09-10
Retry on ConnectionResetError

wjt:
In the past month, this job has failed 7 times. 5 of those were
legitimate failures (the Unity Hub URL broke) and 2 were
ConnectionResetError while reading the Spotify .deb.

It's not great that repository-origin.spotify.com intermittently fails,
but there's nothing we can do about it here.

Make 3 attempts to fetch files, with a 2-second sleep between attempts,
before failing. Use the 'tenacity' library
<https://tenacity.readthedocs.io/> rather than writing the retry loop by
hand.

I also cleaned up some style nits.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@ RUN apt-get update \
       python3-gi \
       python3-github \
       python3-ruamel.yaml \
+      python3-tenacity \
   && apt-get clean \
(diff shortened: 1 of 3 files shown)


andrunko (approved):
lgtm

PR #40, opened 2019-10-21
Clean up imports in tests

wjt:
General cleanup so that the test suite can be run with:

```
python3 -m unittest discover
```

without fiddling with import paths, etc.


andrunko (approved):
LGTM and seems to work fine with local tests.

PR #43, opened 2019-11-13
Add "canonicalize-manifest"  tool

wjt:
This was requested by @hadess at https://github.com/flathub/com.adobe.Flash-Player-Projector/pull/28#issuecomment-553338831 as a way to reformat the JSON/YAML ahead of time so that PRs from this tool are clearer. I guess there is an argument that the tool should produce 2 separate commits, one with any reformatting but no semantic changes, and another with just the actual changes, but this is easier for now.

While I was here, I chucked in a patch that's been lying around in my tree for a few weeks, and some additional debug output from htmlchecker.

### DIFF
--- a/canonicalize-manifest
+++ b/canonicalize-manifest
@@ -1,0 +1,39 @@
+#!/usr/bin/env python3
+#
+# canonicalize-manifest: Reformat a manifest to match the output of the full checker.
+#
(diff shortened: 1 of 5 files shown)


hadess:
Looks good to me.

PR #44, opened 2019-11-19
wrappers: Remove us.zoom.Zoom

barthalion:
Flathub enabled Actions-backed external-data-checker for Zoom [today][1]. Let's see how it goes.

[1]: https://github.com/flathub/us.zoom.Zoom/commit/61980e55fd7e1b9f5e948c964c683bc46f5e7a8d


wjt (approved):
A momentous occasion!

I spotted https://github.com/flathub/flatpak-external-data-checker/commit/76efedea4719bc409587f0999773e2af3e6d4517 – perhaps we should make this configurable.

barthalion:
Probably! I will look into that and submit some PRs when we're sure it works as it should. Thanks!

PR #45, opened 2019-11-28
Flathub integration

barthalion:
Left to do:

- [x] Skip `data` URLs
- [x] Skip loading external files if bigger than `${x}`


src/checker.py
```diff
@@ -77,7 +80,12 @@ def _get_module_data_from_json(self, path, json_data):
             if isinstance(module, str):
                 module_path = os.path.join(os.path.dirname(self._manifest),
                                            module)
-                module = self._read_manifest(module_path)
+                try:
+                    module = self._read_manifest(module_path)
+                except GLib.Error as err:
+                    if err.matches(GLib.quark_from_string('g-file-error-quark'), 4):
```
wjt, line 87:
```suggestion
                    if err.matches(GLib.quark_from_string('g-file-error-quark'), GLib.FileError.NOENT):
```

but I think that if the manifest is a YAML file, this will raise `FileNotFoundError` instead. You could either handle both here, or (maybe neater) make `def read_json_manifest(manifest_path):` convert from the GLib flavour to the Python flavour, or replace:

```
    parser.load_from_file(manifest_path)
```

with
```
with open(manifest_path, "r") as f:
    parser.load_from_data(f.read())
```

src/checker.py
```diff
@@ -77,7 +80,12 @@ def _get_module_data_from_json(self, path, json_data):
...
+                        log.info("Referenced file not found: %s", module)
+                        continue
```
wjt, line 89:
```suggestion
                        continue
                        
                    raise
```

src/lib/utils.py
```diff
@@ -101,6 +101,20 @@ def get_extra_data_info_from_url(url):
     return external_file, data
 
 
+def _check_bwrap():
+    bwrap_cmd = ["bwrap"]
+    for path in ("/usr", "/lib", "/lib64", "/bin", "/proc"):
+        bwrap_cmd.extend(["--ro-bind", path, path])
```
wjt: Might be nice to not have to copy-paste this here and below, but hey ho.

src/lib/utils.py
```diff
@@ -101,6 +101,20 @@ def get_extra_data_info_from_url(url):
...
+
+    bwrap_cmd.extend(["/usr/bin/true"])
+
+    p = subprocess.run(bwrap_cmd)
+    if p.returncode == 0:
+        return True
+
+    return False
```
wjt, line 137:
```suggestion
    logging.warning("bwrap not available")
    return False
```

src/lib/utils.py
```diff
@@ -113,24 +127,27 @@ def extract_appimage_version(basename, data):
...
+                    tmpdir,
+                    tmpdir,
+                    "--die-with-parent",
+                    "--new-session",
+                    "--unshare-all"
+                ]
+             )
+        else:
+            args = []
+            args.extend([appimage_path,"--appimage-extract"])
```
wjt: You aren't actually putting these two elements into the `bwrap` command any more

wjt (changes requested):

src/main.py
```diff
@@ -148,9 +149,11 @@ def open_pr(subject, body, branch):
         log.debug("origin repo not writable; creating fork")
         fork = user.create_fork(origin_repo)
         remote = user.login
-        check_call(("git", "remote", "add", remote, fork.ssh_url))
         repo = fork
 
+    remote_url = f"https://{github_token}:x-oauth-basic@github.com/{repo.full_name}"
+    check_call(("git", "remote", "set-url", remote, remote_url))
```
wjt: I think this will fail on the `remote = user.login` branch since the remote won't've been created.

src/lib/utils.py
```diff
@@ -113,24 +132,24 @@ def extract_appimage_version(basename, data):
...
+                    "--die-with-parent",
+                    "--new-session",
+                    "--unshare-all"
+                ]
+            )
+
+        args.extend([appimage_path,"--appimage-extract"])
+
         log.debug("$ %s", " ".join(args))
+        p = bwrap_run(args, cwd=tmpdir)
```
wjt: Now you'll try to run `bwrap` even on the path where you've determined it's not present :(

wjt: I am a bit concerned about executing arbitrary code, with network access, in an environment that has access to (at least) an API key that can push to this repo. I suppose you could argue that it's the same arbitrary code that the end-user will run on their actual computer anyway…

barthalion: > Now you'll try to run bwrap even on the path where you've determined it's not present :(

Yeah, noticed it too. Fixed.

barthalion: > I am a bit concerned about executing arbitrary code, with network access, in an environment that has access to (at least) an API key that can push to this repo

It's valid concern, but on the other hand, it's not something I can really change. I really want to avoid extending Flathub's buildbot to accommodate anything beyond building applications and basic post-build checks.

I guess I could switch to using https://github.com/refi64/unappimage for extraction.

wjt: Yeah, it's probably quite low risk in practice. Perhaps open an issue for trying to switch to unappimage in futuer?

wjt:
Rebased on master to pick up the CI changes

src/lib/utils.py
wjt: This is now wrong in the `bwrap` case again :(

wjt: ah, fixed by later commit

Issue #42, opened 2019-11-02
Support rewriting scraped URLs


wjt:
Anydesk is checked by scraping their human-readable download page. This links to a URL of the form `https://download.anydesk.com/linux/anydesk-$VERSION-$ARCH.tar.gz`.

When a new version is released, those files disappear, rendering the Flatpak non-installable. However, older (and current) versions remain available at `https://download.anydesk.com/linux/generic-linux/anydesk-$VERSION-$ARCH.tar.gz`.

We could teach htmlchecker to be able to mangle URLs. What I had in mind was something like:

```patch
"x-checker-data": {
    "type": "html",
    "url": "https://anydesk.com/en/downloads/linux",
    "url-pattern": "(https://download.anydesk.com/linux/anydesk-([0-9.]+)-amd64.tar.gz)",
    "version-pattern": "https://download.anydesk.com/linux/anydesk-([0-9.]+)-amd64.tar.gz",
    "url-substitution": ["/linux/", "/linux/generic-linux/"]
}
```

and the checker would just use `str.replace(url_substitution[0], url_substitution[1])`.

An alternative that would work in this case would be to instead scrape the index at https://download.anydesk.com/linux/, find all matches of the `url-pattern`, and take the one with the larger version number.

jakobjakobson13:
Are there any updates on this issue regarding its implementation? This feature would be really useful even in a slimmed down variant that checks only a directory of an ftp server for a newer version.

wjt:
@jakobjakobson13 pull requests very welcome!

wjt:
I think we can now cover the Anydesk case using the feature added in #46.

PR #46, opened 2019-12-03
HTMLChecker: Supports constructing the URL from the matched version

Lctrs:


Lctrs:
WDYT @barthalion ?

wjt:
I think this should be folded into htmlchecker, personally

Lctrs:
It would violate Single Responsibility Principle IMHO.

wjt:
90% of this source file is a copy-paste from `htmlchecker.py` (including the copyright attribution!) which violates Don't Repeat Yourself.

Lctrs:
As of today, yes. But this checker might evolves later. For example, we can add support for modifying the matched version on the fly (like removing all dots in it, etc.) before building the URL with it. Let's not htmlchecker become an `if-else` mess.

barthalion:
@Lctrs I haven't written even 10% of current code base so I'm poor choice for reviewing. I'd prefer to have it in htmlchecker.py for now, or create a class which contains the common code.

Lctrs:
I'm not familiar with Python. Will see what I can do.

Lctrs:
On the plus side, I tested it with one of the flathub repo I maintain and it works well :)

Lctrs:
@wjt @barthalion ready for review.

tests/test_htmlchecker.py
```diff
@@ -56,6 +60,27 @@ def test_check(self):
...
+        checker = ManifestChecker(TEST_MANIFEST_WITH_URL_TEMPLATE)
+        ext_data = checker.check()
+
+        data = self._find_by_filename(ext_data, "xdebug.tar.gz")
+        self.assertIsNotNone(data)
+        self.assertEqual(data.filename, "xdebug.tar.gz")
+        self.assertIsNotNone(data.new_version)
+        self.assertEqual(
+            data.new_version.url,
+            "https://xdebug.org/files/xdebug-2.9.0.tgz",  # noqa: E501
```
wjt: I don't think you need to suppress E501 here

src/checkers/htmlchecker.py
```diff
@@ -71,7 +72,13 @@ def check(self, external_data):
                 "%s had no available version information", external_data.filename
             )
         if not latest_url:
-            log.warning("%s had no available URL", external_data.filename)
+            try:
+                latest_url = Template(external_data.checker_data["url-template"]).substitute(
+                    version=latest_version
+                )
+            except KeyError:
+                pass
```
wjt: You're dropping a legitimate warning here. Perhaps:

```python3
try:
    url_template = external_data.checker_data["url-template"]
except KeyError:
    log.warning("%s had no available URL", external_data.filename)
else:
    latest_url = Template(url_template).substitute(version=latest_version)
```

tests/test_htmlchecker.py
wjt: Will this test not start failing at some point in the future when xdebug makes a new release?

wjt (changes requested):
This is great. I think it could also resolve #42: Anydesk could scrape the version number from the URL, then build the stable URL from it.

Please could you rebase to drop the patch from #47? I also have a couple of points about the checker and its test.

tests/test_htmlchecker.py
Lctrs: > Will this test not start failing at some point in the future when xdebug makes a new release?

No because I specifically target this release in tests/org.xdebug.Xdebug.json.

Lctrs:
@wjt Rebased and comments adressed.

wjt (approved):

tests/test_htmlchecker.py
wjt: Gotcha!

PR #47, opened 2019-12-03
Add new JetBrains checker

Lctrs:
Missing tests

### DIFF
--- a/src/checkers/jetbrainschecker.py
+++ b/src/checkers/jetbrainschecker.py
@@ -1,0 +1,44 @@
+import datetime
+import json
+import logging
+import urllib.error
+import urllib.parse
+import urllib.request
+
+from src.lib.externaldata import ExternalFile, Checker
+
+log = logging.getLogger(__name__)
+
+
+def _should_check(external_data):
+    return external_data.checker_data.get("type") == "jetbrains"
+
+
+class JetBrainsChecker(Checker):
+    def check(self, external_data):
+        if not _should_check(external_data):
+            log.debug("%s is not a jetbrains type ext data", external_data.filename)
+            return
+
+        code = external_data.checker_data["code"]
+        release_type = external_data.checker_data.get("release-type", "release")
+
+        url = f"https://data.services.jetbrains.com/products/releases?code={code}&latest=true&type={release_type}"
+
+        log.debug("Getting extra data info from %s; may take a while", url)
+        resp = urllib.request.urlopen(url)
+        data = json.load(resp)[code][0]
+        release = data["downloads"]["linux"]
+
+        checksum = urllib.request.urlopen(release["checksumLink"]).read().decode("utf-8").split(" ")[0]
+
+        new_version = ExternalFile(
+            release["link"],
+            checksum,
+            release["size"],
+            data["version"],
+            datetime.datetime.strptime(data["date"], "%Y-%m-%d")
+        )
+
+        if not external_data.current_version.matches(new_version):
+            external_data.new_version = new_version
--- a/tests/test_jetbrainschecker.py
+++ b/tests/test_jetbrainschecker.py
@@ -1,0 +1,46 @@
+import os
+import unittest
+
+from src.checker import ManifestChecker
+from src.lib.utils import init_logging
+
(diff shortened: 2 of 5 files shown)


Lctrs:
Kudos to @x80486 for the original code :)

Lctrs:
Ping @FakeShemp and @johanfleury. As maintainers of flatpak wrappers for some Jetbrains products, you might find some interest in this.

Lctrs:
@wjt ready for review :pray:

Lctrs:
Test result is here : https://github.com/Lctrs/flatpak-external-data-checker/commit/68ec6a40b6306b4e89eb85c1209f19e46103d963/checks?check_suite_id=343482923

src/checkers/jetbrainschecker.py
```diff
@@ -0,0 +1,45 @@
...
+
+class JetBrainsChecker(Checker):
+    def check(self, external_data):
+        if not _should_check(external_data):
+            log.debug("%s is not a jetbrains type ext data", external_data.filename)
+            return
+
+        code = external_data.checker_data.get("code")
+
+        assert code is not None
```
wjt: You can just write

```python3
code = external_data.checker_data["code"]
```

src/checkers/jetbrainschecker.py
```diff
@@ -0,0 +1,45 @@
...
+
+        url = Template("https://data.services.jetbrains.com/products/releases?code=$code&latest=true&type=$type")\
+            .substitute(
+                code=code, type=external_data.checker_data.get("release-type", "release")
+            )
```
wjt:
```python3
release_type = external_data.checker_data.get("release-type", "release")
url = f"https://data.services.jetbrains.com/products/releases?code={code}&latest=true&type={type}"
```

src/checkers/jetbrainschecker.py
```diff
@@ -0,0 +1,45 @@
...
+        data = json.load(resp)[code][0]
+        release = data["downloads"]["linux"]
+
+        checksum = urllib.request.urlopen(release["checksumLink"]).read().decode("utf-8").split(" ")[0]
+
+        new_version = ExternalFile(
+            release["link"], checksum, release["size"], data["version"], data["date"]
+        )
+
+        external_data.state = ExternalData.State.VALID
```
wjt: This is meant to refer to the state of the _current_ version, not of the new version. I think you can just drop this line.

tests/test_jetbrainschecker.py
```diff
@@ -0,0 +1,97 @@
...
+            b"dbb7bf79750b476d2c96e76be6c6082d63c2addffd2694d4d9fa9d6db01162a2 *PhpStorm-2019.3.tar.gz",
+            200
+        )
+
+    return MockResponse(None, 404)
+
+
+class TestJetBrainsChecker(unittest.TestCase):
+    def setUp(self):
+        logging.basicConfig(level=logging.DEBUG)
```
wjt: use init_logging() from utils (see c7f0a7cd1591a4d9b03d78f751566a7942f1935f)

wjt (changes requested):
This is a great idea, thanks for the contribution! Just a few bits and bobs.

src/checkers/jetbrainschecker.py
Lctrs: ~~Looks like this line is needed ? :thinking: The test is failing now.~~

src/checkers/jetbrainschecker.py
Lctrs: I'm not a Python developer (in fact, those are my first Python lines :tada:), but I read on this [blog post](https://realpython.com/python-string-formatting/) that String interpolation in Python are not safe from user input, while Template Strings are. I'm concerned that a malicious user can just get @flathubbot's token access in the log of the action for example.

barthalion: Technically speaking, I can think of multiple ways a malicious user could already take and none includes string interpolation… So I wouldn't worry too much.

Lctrs: Okay done.

wjt (approved):
LGTM, thanks!

PR #55, opened 2019-12-05
Merge PRs automatically if enabled in flathub.json

barthalion:


wjt (commented):
I think we'd want to make this trigger under slightly different circumstances. Rather than being purely configuration controlled, we should only automerge in the case where the app is no longer installable due to 1 or more current URLs being dead or not matching the checksum. (In that case, merging a broken automated change can't make the situation any worse – at least in theory.)

Having done that, I'm not sure whether we'd still want a configuration flag – probably yes, but maybe not?

barthalion:
IMHO we still want both – VueScan is a good example of application with hyper-active vendor. There are days when checksum changes 4-5 times.

That being said, I'm not sure how to cleanly implement automatic merge for broken URLs at the moment so I will handle that separately in January.

wjt:
https://github.com/flathub/flatpak-external-data-checker/blob/86d0a601a24e64989a4260b2bb80700baeb2da95/src/lib/externaldata.py#L60-L65

This enum is meant to allow us to distinguish between "current URL is valid" and "current URL is broken", orthogonally to "an update is available".

barthalion:
Thanks for the tip, latest commit adds enabling automatic merge for modules with broken URL and new version available. I haven't tested it yet; my guess is that we should respect maintainers' decision if they explicitly set `automerge-flathubbot-prs` to false.

src/main.py
```diff
@@ -161,15 +167,44 @@ def open_pr(subject, body, branch):
     base = "master"
     head = "{}:{}".format(repo.owner.login, branch)
     pr_message = ((body or "") + "\n\n" + DISCLAIMER).strip()
-    # Include closed PRs – if the maintainer has closed our last PR, we don't want to
+
+    if os.path.isfile("flathub.json"):
```
barthalion, line 172: Now thinking about it, I wonder if it should kept configurable at all. I can't think of a good reason why someone would want to disable auto-merge for invalid URLs.

src/main.py
```diff
@@ -130,9 +131,14 @@ def commit_changes(changes):
 )
 
 
-def open_pr(subject, body, branch):
+def open_pr(subject, body, branch, manifest_checker=None):
+    try:
+        github_token = os.environ["GITHUB_TOKEN"]
+    except KeyError:
+        log.error("GITHUB_TOKEN environment variable is not set")
+        return
```
wjt: If this is an error, the script should not exit(0).

src/main.py
```diff
@@ -161,16 +167,46 @@ def open_pr(subject, body, branch):
...
-    # open another one.
-    for pr in origin_repo.get_pulls(state="all", base=base, head=head):
+
+    if os.path.isfile("flathub.json"):
+        with open("flathub.json") as f:
+            repocfg = json.load(f)
+            automerge = repocfg.get("automerge-flathubbot-prs")
+
+    # Enable automatic merge if external data is broken unless explicitly disabled
+    if automerge is not False and manifest_checker:
```
wjt, line 178:
```suggestion
    if automerge and manifest_checker:
```

src/main.py
```diff
@@ -161,16 +167,46 @@ def open_pr(subject, body, branch):
...
+    if origin_repo.permissions.push:
+        open_prs = [pr for pr in prs if pr.state == "open"]
+        for pr in open_prs:
+            log.info("Found open PR: %s", pr.html_url)
+
+            if automerge:
+                pr_commit = pr.head.repo.get_commit(pr.head.sha)
+                if pr_commit.get_combined_status().state == "success" and pr.mergeable:
+                    log.info("PR passed CI and is mergeable, merging", pr.html_url)
+                    pr.merge(merge_method='rebase')
```
wjt: It might be nice if the bot left a note on the PR to explain why it is doing this.

src/main.py
barthalion: That would make no sense. It's supposed to enable automerge if it's not explicitly disabled (thus `is not False`) in flathub.json and some source is broken.

src/main.py
barthalion: Done.

src/main.py
barthalion: Done.

src/main.py
```diff
@@ -161,16 +172,48 @@ def open_pr(subject, body, branch):
...
+    for pr in closed_prs:
         log.info(
             "Found existing %s PR: %s",
             "merged" if pr.is_merged() else pr.state,
             pr.html_url,
         )
         return
 
+    if origin_repo.permissions.push:
+        open_prs = [pr for pr in prs if pr.state == "open"]
```
wjt: By moving the check for open PRs into this block, I think the behaviour for repos where the script _can't_ push to the origin repo will be changed: if a PR is open for the branch, it will still try and fail to create another PR for the same branch.

I think the behaviour we want, logically, is:

* if there is at least one PR for this branch:
  - if the source is currently broken, and the PR is open, and CI passed, and automerge is enabled, and it can be merged, then merge it
  - otherwise, return
* otherwise, push the branch and open a PR

wjt:
As discussed on https://github.com/flathub/com.dropbox.Client/pull/80#issuecomment-571775406, urlchecker currently incorrectly sets BROKEN even if the old URL is still reachable. We'd need to fix this before putting this branch into production … oops!

wjt:
@barthalion this branch is already in production, right? What other code is unmerged to `master` in this repo and yet run in prod? Just this?

barthalion:
Yeah, we are currently using this branch. I'll get back to resolving remaining issues tomorrow.

src/main.py
barthalion: As GitHub very helpfully says `outdated` and I don't remember what the code was doing before it was labeled so, the current behavior is:

1. Check if there's newer data available.
2. Check if old data is 
3. If both conditions are met, enable automerge.

Then:

- Find open PRs.
- Check if have permissions to push and automerge is enabled.
  - Return if not.
  - Check if CI passed for PR and is mergeable.
    - Merge and return.

So if I understand correctly, the remaining issue here is that the function should return if found an open PR but have no push permissions.

barthalion: Ah, yes, I can't read. That's already the case.

barthalion:
https://github.com/flathub/flatpak-external-data-checker/blob/aa9095289e3fb7812ff74e7f1037bdee5a967a4c/src/checkers/urlchecker.py#L108

I assume it should be set to `VALID` here then?

https://github.com/flathub/flatpak-external-data-checker/blob/aa9095289e3fb7812ff74e7f1037bdee5a967a4c/src/checkers/urlchecker.py#L85

This is out of scope for this PR, but I think we need to reverse the conditions here and check `is_rotating_url` first and then fallback to AppImage version extraction.

PR #58, opened 2019-12-10
Support for nested modules

gasinvein:
Handle sub-modules of other modules, e.g. cases like this:
```yaml
id: tld.domain.App
modules:
  - name: the-app
    sources:
      - ...
    modules:
      - name: some-dependency
        sources:
          - ...
```
Currently sub-modules are ignored.

### DIFF
--- a/src/checker.py
+++ b/src/checker.py
@@ -67,9 +67,11 @@ class ManifestChecker:
         dump_manifest(contents, path)
 
     def _collect_external_data(self, path, json_data):
-        modules = json_data.get('modules', [])
+        modules = json_data.get('modules')
         if modules is None:
-            log.error("Failed to deserialize \"modules\" in %s", path)
+            return
+        elif not isinstance(modules, list):
+            log.warning("\"modules\" in %s is not a list", path)
             return
         for module in modules:
             if isinstance(module, str):
--- a/src/checker.py
+++ b/src/checker.py
@@ -69,8 +69,9 @@ class ManifestChecker:
     def _collect_external_data(self, path, json_data):
         for module in json_data.get('modules', []):
             if isinstance(module, str):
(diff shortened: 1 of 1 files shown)


src/checker.py
```diff
@@ -67,7 +67,11 @@ def _dump_manifest(self, path):
         dump_manifest(contents, path)
 
     def _collect_external_data(self, path, json_data):
-        for module in json_data.get('modules', []):
+        modules = json_data.get('modules', [])
+        if modules is None:
+            log.error("Failed to deserialize \"modules\" in %s", path)
+            return
+        for module in modules:
```
wjt, line 76: `error` definitely seems too high severity here, and the proposed message is not correct – it was successfully deserialized, as `None`. I personally wouldn't bother logging anything. I suggest:

```python3
modules = json_data.get("modules")
if not modules:
    return
for module in modules:
```

wjt (changes requested):
Great catch! One change proposed. A test would be nice but not essential.

src/checker.py
gasinvein: Maybe check if `modules` is a `list`? Any other type probably should be ignored.

gasinvein: Would this be fine?
```python
        modules = json_data.get('modules')
        if modules is None:
            return
        elif not isinstance(modules, list):
            log.warning("Failed to deserialize \"modules\" in %s", path)
            return
```

wjt: Sure, I'd prefer something like:

```python3
            log.warning('"modules" in %s is not a list', path)
```

(or with the quotes the other way around, to taste)

wjt (approved):

wjt:
Thank you!

gasinvein:
No problem. If I'll manage to create a test-case for this, will open a new PR.

PR #60, opened 2019-12-18
htmlchecker: Handle multiple matches

gasinvein:
Single HTML page can list multiple program versions to download. In this case, pick the latest one, not the first match (which is likely to be the oldest version, given they are listed ascending).
This makes specifying exactly one group in pattern mandatory.

### DIFF
--- a/src/checkers/htmlchecker.py
+++ b/src/checkers/htmlchecker.py
@@ -41,12 +41,15 @@ def get_latest(checker_data, pattern_name, html):
     except KeyError:
         return None
 
-    m = pattern.search(html)
-    if m is None:
+    m = pattern.findall(html)
+    if not m:
         log.debug("%s %s did not match", pattern_name, pattern)
         return None
+    if len(m) > 1:
+        log.debug("%s %s matched multiple times, picking last match", pattern_name, pattern)
 
-    result = m.group(1)
+    result = m[-1]
+    assert isinstance(result, str)
     log.debug("%s %s matched: %s", pattern_name, pattern, result)
     return result
 
--- a/src/checkers/htmlchecker.py
+++ b/src/checkers/htmlchecker.py
@@ -42,7 +42,8 @@ def get_latest(checker_data, pattern_name, html):
     except KeyError:
         return None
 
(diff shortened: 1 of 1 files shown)


src/checkers/htmlchecker.py
```diff
@@ -40,12 +41,18 @@ def get_latest(checker_data, pattern_name, html):
...
+    assert pattern.groups == 1
+
+    m = pattern.findall(html)
+    if not m:
         log.debug("%s %s did not match", pattern_name, pattern)
         return None
+    if len(m) == 1:
+        result = m[0]
+    else:
+        result = max(m, key=LooseVersion)
```
gasinvein, line 56: I'm not sure if using `distutils.version.LooseVersion` is the best way to compare versions, but at least it's already there.

src/checkers/htmlchecker.py
```diff
@@ -40,12 +41,18 @@ def get_latest(checker_data, pattern_name, html):
     except KeyError:
         return None
 
-    m = pattern.search(html)
-    if m is None:
+    assert pattern.groups == 1
```
wjt:
```suggestion
    if pattern.groups != 1:
        raise ValueError(f"{pattern_name} {pattern} does not contain exactly 1 match group")
```

src/checkers/htmlchecker.py
wjt: In the case of `url-pattern` the results will be URLs, not version strings, though in practice I guess LooseVersion will probably do the right thing.

wjt (changes requested):
One relatively minor suggestion. I think exercising this code path in tests would be nice, but not essential.

src/checkers/htmlchecker.py
gasinvein: `LooseVersion` seems to handle any string just fine, but I'm not sure how reliable it would be to use it this way.

wjt:
Looks good, but please can you rebase and resolve the conflicts?

gasinvein:
@wjt Rebased.

PR #61, opened 2019-12-19
Debianrepochecker: also get arch from flathub.json

Lctrs:
If only one arch is defined in flathub.json, use this one.

Following comment made by @barthalion in https://github.com/flathub/flathub/pull/1292#discussion_r357015490

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -24,17 +24,16 @@
 
 import argparse
 import contextlib
+import json
 import logging
 import os
 import subprocess
-import sys
 
 from github import Github
 
-from src.lib.utils import parse_github_url, init_logging
-from src.lib.externaldata import ExternalData
 from src import checker
-
+from src.lib.externaldata import ExternalData
+from src.lib.utils import parse_github_url, init_logging
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


src/main.py
```diff
@@ -183,6 +182,14 @@ def open_pr(subject, body, branch):
     log.info("Opened pull request %s", pr.html_url)
 
 
+def load_flathubconfig(path):
+    if not os.path.isfile(path + "/flathub.json"):
+        return None
+
+    with open(path + "/flathub.json") as f:
+        return json.load(f)
```
wjt: How about:

```python3
try:
    with open(os.path.join(path, "flathub.json"), "r") as f:
        return json.load(f)
except FileNotFoundError:
    return {}
```

wjt:
I've got to say that I don't agree with the comment at https://github.com/flathub/flathub/pull/1292#discussion_r357015490 – I think it's correct for the manifest to be self-contained and not be reliant on information from `flathub.json` – but nonetheless this feature seems useful since evidently others disagree :-)

src/main.py
Lctrs: Adressed.

Lctrs:
I have no strong opinion on it. Let's wait for @barthalion input.

barthalion:
Yeah, I don't think using flathub.json is fine either. I'd rather add `arch` field into `x-checker-data` and keep `only-arches` as a fallback only if it's unavailable here. (Yes, I'm not a fan of mixing flatpak-builder and external tooling metadata.)

Lctrs:
Okay. Closing then.

PR #63, opened 2019-12-23
Add support for updating metainfo file

Lctrs:


Lctrs:
Failure is unrelated.

cc @wjt

PR #64, opened 2019-12-23
lib/externaldata: Look for arches in checker_data, default to x86_64

barthalion:
Flathub apps often limit target architectures with only-arches in
flathub.json, making the manifest counterpart redundant. Look for
arches in x-checker-data first, then in only-arches, finally default
to x86_64.

### DIFF
--- a/src/lib/externaldata.py
+++ b/src/lib/externaldata.py
@@ -113,14 +113,7 @@ class ExternalDataSource(ExternalData):
         sha256sum = source.get('sha256')
         size = source.get('size', -1)
         checker_data = source.get('x-checker-data', {})
-
-        # TODO: Switch to walrus operator when we switch to Python 3.8
-        if checker_data.get('arches'):
-            arches = checker_data.get('arches')
-        elif source.get('only-arches'):
-            arches = source.get('only-arches')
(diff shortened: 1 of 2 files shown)


src/lib/externaldata.py
```diff
@@ -110,13 +110,22 @@ def __init__(self, source_path, source, sources, data_type, url):
...
-        checker_data = source.get('x-checker-data')
+        checker_data = source.get('x-checker-data', {})
+
+        # TODO: Switch to walrus operator when we switch to Python 3.8
+        if checker_data.get('arches'):
+            arches = checker_data.get('arches')
+        elif source.get('only-arches'):
+            arches = source.get('only-arches')
+        else:
+            arches = ["x86_64"]
```
wjt: How about something like:

```python3
arches = checker_data.get("arches") or source.get("only-arches") or ["x86_64"]
```

barthalion: Done.

wjt:
Sorry, I didn't realise that _Squash & Merge_ re-attributes the commit to me.

barthalion:
No worries. I've had no idea either.

PR #65, opened 2019-12-23
Use black to check for proper code formatting and reformat existing code

barthalion:


src/checkers/firefoxchecker.py
```diff
@@ -39,44 +39,50 @@
...
 class FirefoxChecker(Checker):
-    FIREFOX_RELEASE_INFO_URL = 'https://product-details.mozilla.org/1.0/firefox_versions.json'  # noqa: E501
-    FIREFOX_ARCHIVE_BASE_URL = 'https://archive.mozilla.org/pub/firefox/releases/{version}/'  # noqa: E501
-    FIREFOX_PLATFORM = 'linux-x86_64'
-    CHECKER_BROWSER_SOURCE_FILENAME = 'firefox.tar.bz2'
-    CHECKER_BROWSER_FILENAME_MATCH = r'firefox-.+\.tar\.bz2$'
-    CHECKER_BROWSER_DEFAULT_LANGUAGE = 'en-US'
-    CHECKER_TRANSLATIONS_FILENAME_MATCH = r'.+\.xpi$'
+    FIREFOX_RELEASE_INFO_URL = (
+        "https://product-details.mozilla.org/1.0/firefox_versions.json"  # noqa: E501
```
wjt:
```suggestion
        "https://product-details.mozilla.org/1.0/firefox_versions.json"
```

at which point it probably fits on one line again

src/checkers/firefoxchecker.py
```diff
@@ -39,44 +39,50 @@
...
+    )
+    FIREFOX_ARCHIVE_BASE_URL = (
+        "https://archive.mozilla.org/pub/firefox/releases/{version}/"  # noqa: E501
```
wjt:
```suggestion
        "https://archive.mozilla.org/pub/firefox/releases/{version}/"
```

src/checkers/firefoxchecker.py
```diff
@@ -39,44 +39,50 @@
...
             if re.match(self.CHECKER_BROWSER_SOURCE_FILENAME, data.filename):
                 browser_data = data
                 break
 
         if not browser_data:
-            log.warning('Unable to find browser source with filename %s',
-                        module_data.name,
-                        self.CHECKER_BROWSER_SOURCE_FILENAME)
+            log.warning(
+                "Unable to find browser source with filename %s",
```
wjt: The reformatting here is correct but it has shown that the log message is wrong – it only has one `%s` placeholder.

src/main.py
```diff
@@ -174,34 +176,39 @@ def open_pr(subject, body, branch):
...
+    parser.add_argument(
+        "manifest", help="Flatpak manifest to check", type=os.path.abspath
+    )
+    parser.add_argument(
+        "-v", "--verbose", help="Print debug messages", action="store_true"
+    )
+    parser.add_argument(
+        "--update",
+        help="Update manifest(s) to refer to new versions of "
+        "external data - also open PRs for changes unless "
```
wjt, line 191: I really, really loathe this aspect of `black`'s formatting, but still not enough to do anything about it like “file a bug”.

wjt (changes requested):
OK! I've been doing this piecemeal (as you may be able to guess from `.flake8`) but I guess we should bite this bullet at some point. That said:

src/checkers/firefoxchecker.py
barthalion: I might be wrong but original code seems to have only one placeholder as well.

wjt: Yes, I know. Up to you if you fancy fixing it.

src/main.py
barthalion: Yeah, I'm not particularly happy about it either, but it hand-waves the entire class of code formatting issues like `gofmt` does. I can live with opinionated tool that makes me not think about it even if I dislike some things.

src/checkers/firefoxchecker.py
barthalion: Will look into that tomorrow.

src/checkers/firefoxchecker.py
barthalion: Ah, sure, misunderstood your comment.

src/checkers/firefoxchecker.py
barthalion: It doesn't seem to affect the way Black formats it.

barthalion:
@wjt Rebased on master and addressed your comments.

wjt (approved):

PR #66, opened 2019-12-27
ci: don't fail if Coveralls token is missing

wjt:
This happens when a PR is submitted by someone without write access to
the repo.


wjt:
I tried using [a condition](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepsif) but the `secrets` object is apparently not defined there.

I also considered using a bash condition but ultimately I don't really care if coveralls is broken.

PR #68, opened 2020-01-06
main: Use pygit2 to determine current branch

barthalion:
Some Flathub repositories use different branch than master as the
default.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,7 @@ RUN apt-get update \
       python3-gi \
       python3-github \
       python3-pip \
+      python3-pygit2 \
       python3-ruamel.yaml \
       python3-setuptools \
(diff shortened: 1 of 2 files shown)


barthalion:
cc @nanonyme

nanonyme:
Looks good to me. If you pull in pygit2, it possibly makes sense to implement the git tracking later using the same tool.

wjt:
I reverted this because it doesn't work – see stack trace on #70. That'll teach me to merge PRs from a tablet without testing myself!

barthalion:
Ah, sorry, I should have tested it more carefully too. I'll resubmit the change without shadowing existing variables.

If we were popular, there would be Rust rewrite underway which would catch such mistakes!

PR #79, opened 2020-03-25
urlchecker: Strip query string from redirect target

### LINKED ISSUE #78: zoom gets loads of pull requests
e.g. https://github.com/flathub/us.zoom.Zoom/pull/110
this is quite annoying :-/

wjt:
This script is spamming the Zoom repo with patches like this:

```patch
-"url": "https://d11yldzmag5yn.cloudfront.net/prod/3.5.372466.0322/zoom_x86_64.tar.xz",
+"url": "https://d11yldzmag5yn.cloudfront.net/prod/3.5.372466.0322/zoom_x86_64.tar.xz?_x_zm_rtaid=muDd1uOqSZ-xUScZF698QQ.1585134521724.21e5ab14908b2121f5ed53882df91cb9&_x_zm_rhtaid=732",
```

This appears to be a marketing tag, and as such changes every time the
bot scrapes the homepage.

To avoid this, strip all query strings when following a rotating-url.

In principle, there might be situations where the query string has to be
preserved, so this would need to be configurable. In practice, I expect
that only the test case changed in this patch will be affected.

Fixes #78

### DIFF
--- a/src/lib/utils.py
+++ b/src/lib/utils.py
@@ -69,6 +69,16 @@ def _extract_timestamp(info):
         return dt.datetime.now()  # what else can we do?
 
 
+def strip_query(url):
+    """Strips the query string from the given URL, if any."""
+    parts = urllib.parse.urlparse(url)
+    if not parts.query:
+        return url
+    stripped = urllib.parse.urlunparse(parts._replace(query=""))
+    log.debug("Normalised %s to %s", url, stripped)
+    return stripped
+
+
 def get_timestamp_from_url(url):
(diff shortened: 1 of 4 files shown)


barthalion (approved):

jurf:
Could we have an option to re-enable it back? E.g. TLauncher needs the query strings for the download: https://github.com/flathub/org.tlauncher.TLauncher/pull/1

wjt:
Doh, I guess this was wrong:

> In practice, I expect
> that only the test case changed in this patch will be affected.

wjt:
Could you file an issue for that?

PR #80, opened 2020-04-09
Add checker for snapcraft packages

FakeShemp:
<s>This currently doesn't work, but I want some feedback before fixing it.

Snap uses sha3-384 for hashes. Afaik, this is not supported in a Flatpak. How should this be handled? Add support for sha3-384 in Flatpak or try to get another hash some other way?</s>

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,16 +193,16 @@ for [JetBrains](https://www.jetbrains.com/) products:
 }
 ```
 
-#### SnapCraft checker
+#### Snapcraft checker
 
 Special checker that will check for available updates
-for [SnapCraft](https://snapcraft.io/) packages:
+for [Snapcraft](https://snapcraft.io/) packages:
 
 ```json
 "x-checker-data": {
     "type": "snapcraft",
     "name": "PACKAGE-NAME",
-    "release": "stable, beta, or any other tag the project uses"
+    "channel": "stable, beta, or any other tag the project uses"
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -193,6 +193,20 @@ for [JetBrains](https://www.jetbrains.com/) products:
 }
 ```
 
+#### SnapCraft checker
+
+Special checker that will check for available updates
+for [SnapCraft](https://snapcraft.io/) packages:
+
+```json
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -193,6 +193,20 @@ for [JetBrains](https://www.jetbrains.com/) products:
 }
 ```
 
+#### SnapCraft checker
```
barthalion: Snapcraft, not SnapCraft.

README.md
```diff
@@ -193,6 +193,20 @@ for [JetBrains](https://www.jetbrains.com/) products:
...
+
+Special checker that will check for available updates
+for [SnapCraft](https://snapcraft.io/) packages:
+
+```json
+"x-checker-data": {
+    "type": "snapcraft",
+    "name": "PACKAGE-NAME",
+    "release": "stable, beta, or any other tag the project uses"
```
barthalion: This should be `channel`, at least according to snapcraft's dictionary.

barthalion:
Seems flatpak-builder supports sha384. Is it the same thing that sha3-384?

FakeShemp:
@barthalion Unfortunately not. That would be sha2-384.

barthalion:
> @barthalion Unfortunately not. That would be sha2-384.

I don't think there's a good way to implement it in flatpak-builder, it's not supported by glib either if I see correctly. The checker should check if the checksum matches and rewrite it to sha256 if yes.

FakeShemp:
Is this stuck at something special I can fix?

tests/test_snapcraftchecker.py
```diff
@@ -0,0 +1,44 @@
+import os
+import unittest
+
+from src.checker import ManifestChecker
+from src.lib.utils import init_logging
+
+TEST_MANIFEST = os.path.join(os.path.dirname(__file__), "com.nordpass.NordPass.yaml")
+
+
+class TestJetBrainsChecker(unittest.TestCase):
```
barthalion: Should be TestSnapctaftChecker.

barthalion:
Sorry, I forgot about it. Just one minor thing.

PR #81, opened 2020-04-09
Revert "debianrepochecker: strip Debian revision from package"

wjt:
This reverts commit fe82ed9d6a960fe06ead892d0a4d571f26dafd0c.

My only justification for that behaviour was “I think the -1 is ugly”.
Over on https://github.com/flathub/com.visualstudio.code.insiders/pull/1
we have a case where the package revision matters, so let's leave it in
place.

(Of course it could be made configurable but I'd prefer to keep this
simple.)

### DIFF
--- a/src/checkers/debianrepochecker.py
+++ b/src/checkers/debianrepochecker.py
@@ -117,19 +117,11 @@ class DebianRepoChecker(Checker):
         with self._load_repo(root, dist, component, arch) as cache:
             package = cache[package_name]
             candidate = package.candidate
-            # http://www.fifi.org/doc/debian-policy/policy.html/ch-versions.html
-            #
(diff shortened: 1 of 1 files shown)


fooishbar:
Maybe a stupid heuristic like `only stripping if n < 100` is enough to differentiate the cases from which the -1 or -0.1 or whatever is ugly and unnecessary, vs. the ones who actually apply a timestamp or a Git revision? Or it could be configurable indeed. Shrug.

fooishbar:
but cf. flathub/com.visualstudio.code.insiders#2 which is also an update to 1.45.0, distinct from flathub/com.visualstudio.code.insiders#1 updating ... also to 1.45.0.

PR #84, opened 2020-05-26
Only strip query parameters beginning with underscore

### LINKED ISSUE #83: Make stripping query string configurable
5cc21f27dc0e8b94a184bd62ed214241ee211056 unceremoniously strips all query strings from download URLs. I wrote:

> In principle, there might be situations where the query string has to be
> preserved, so this would need to be configurable. In practice, I expect
> that only the test case changed in this patch will be affected.

I was wrong! Per https://github.com/flathub/flatpak-external-data-checker/pull/79#issuecomment-634289305 / https://github.com/flathub/org.tlauncher.TLauncher/pull/1 in some cases the query string is required.

wjt:
This behaviour was introduced to strip tracking junk from Zoom URLs but
caused problems in a case where query parameters are necessary.

Be a little less ruthless by only stripping parameters whose names begin
with underscores, as a heuristic for "is tracking-ish".

Fixes #83

### DIFF
--- a/src/lib/utils.py
+++ b/src/lib/utils.py
@@ -70,11 +70,16 @@ def _extract_timestamp(info):
 
 
 def strip_query(url):
-    """Strips the query string from the given URL, if any."""
+    """Sanitizes the query string from the given URL, if any. Parameters whose
+    names begin with an underscore are assumed to be tracking identifiers and
+    are removed."""
     parts = urllib.parse.urlparse(url)
     if not parts.query:
         return url
-    stripped = urllib.parse.urlunparse(parts._replace(query=""))
+    qsl = urllib.parse.parse_qsl(parts.query)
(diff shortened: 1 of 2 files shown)


src/lib/utils.py
```diff
@@ -70,11 +70,16 @@ def _extract_timestamp(info):
 
 
 def strip_query(url):
-    """Strips the query string from the given URL, if any."""
+    """Sanitizes the query string from the given URL, if any. Parameters whose
+    names begin with an underscore are assumed to be tracking identifiers and]
```
jurf: Stray typo I assume.
```suggestion
    names begin with an underscore are assumed to be tracking identifiers and
```

wjt: Thanks!

jurf:
Thanks for the work!

PR #87, opened 2020-06-17
checker: Only update appdata version if last source changed

### LINKED ISSUE #86: Incorrect updates to appdata version
There is a comment in the code where the appdata version is updated:

https://github.com/flathub/flatpak-external-data-checker/blob/05a99d4cb6aaa924b7b03f31389e6eebde19af77/src/checker.py#L263-L267

The assumption described isn't always a good assumption. For example, a project might bundle a couple versioned dependencies as well. If one of those dependencies is updated but the main application source is not updated, the checker will insert the dependency's version into the appdata file.

I think a better behavior would be for the code to check if the last item in `datas` had a new version and insert that version into the appdata if so. The assumption then is that the last checked source is the one that sets the version for the flatpak rather than the last source that changed for this specific check.

wjt:
@willsALMANJ suggested this improved heuristic on #86. It seems plausible to me, and the tests pass.

Fixes #86.

### DIFF
--- a/src/checker.py
+++ b/src/checker.py
@@ -228,8 +228,6 @@ class ManifestChecker:
         ]
 
     def _update_manifest(self, path, datas, changes):
-        last_update = None
-
         for data in datas:
             if (
                 data.new_version is None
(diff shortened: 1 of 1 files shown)


wshanks:
Just to give a data point on this -- I just discovered that I actually got bitten by #86 when I merged https://github.com/flathub/de.klayout.KLayout/pull/10 without noticing that it changed my app's version number to that of an updated dependency. I just noticed it now when it led to the next real update of the app failing here: https://github.com/flathub/de.klayout.KLayout/pull/11

PR #90, opened 2020-09-20
Add JSON checker

gasinvein:
This is just like HTML checker, but uses `jq` queries on json urls instead of  regular expressions on html pages.

Usage example (for telegram-desktop, source from github releases in attachment):
```json
"type": "archive",
"url": "https://github.com/telegramdesktop/tdesktop/releases/download/v2.3.2/tdesktop-2.3.2-full.tar.gz",
"sha256": "292631bcac4b30f778879ecd2cebf4c6f569ab5be01230c8e62b924b4211a259",
"x-checker-data": {
    "type": "json",
    "url": "https://api.github.com/repos/telegramdesktop/tdesktop/releases/latest",
    "version-query": ".tag_name",
    "url-query": ".assets[] | select(.label==\"Source code (tar.gz, full)\") | .browser_download_url"
}
```

**Warning**:  running queries from untrusted sources poses a security risk. We must either sanitize query before passing it to `jq`, or run the whole thing in a sandbox.
Current implementation executes `jq` in `bwrap` sandbox, yet I'm not sure if this is enough.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,7 @@ RUN apt-get update \
       python3-toml \
       python3-pyelftools \
       squashfs-tools \
+      jq \
   && apt-get clean \
   && rmdir /var/cache/apt/archives/partial
 
--- a/README.md
+++ b/README.md
@@ -114,6 +114,25 @@ the retrieved version:
 }
 ```
 
+### JSON checker
+
+The JSON checker allows using [jq](https://stedolan.github.io/jq/) to query
+JSON data with arbitrary schema to get version and download url.
+
+To use the **JSONChecker**, specify JSON data URL, version query and url query
+(you can use `$version` variable got from the version query in url query):
+
+```json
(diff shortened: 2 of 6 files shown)


gasinvein:
Looks like test if failing due to postmanlabs/httpbin#617

gasinvein:
@wjt Don't mind reviewing this? It seems to work pretty good with Github/Gitlab APIs, but there are two questions:
- [ ] Is `bwrap` sufficient security boundary? (assuming we need one between the user data and f-e-d-c process)
- [ ] Should the boundary be mandatory? E.g. if bwrap check fails - should we continue without it?

gasinvein:
I've removed python-based URL template if flavor of using only JQ for constructing URLs (version got with `version-query` is passed as `$version` variable to `url-query`).

barthalion (approved):

Issue #91, opened 2020-09-23
Data checker fails on manifest for org.gnome.Chess


Zlopez:
I tried to use the data checker for [org.gnome.Chess](https://github.com/flathub/org.gnome.Chess/blob/master/org.gnome.Chess.json) and got 
```
Traceback (most recent call last):
  File "/home/zlopez/git/flatpak-external-data-checker/flatpak-external-data-checker", line 30, in <module>
    main()
  File "/home/zlopez/git/flatpak-external-data-checker/src/main.py", line 258, in main
    manifest_checker.check(filter_type)
  File "/home/zlopez/git/flatpak-external-data-checker/src/checker.py", line 196, in check
    checker.check(data)
  File "/home/zlopez/git/flatpak-external-data-checker/src/checkers/htmlchecker.py", line 81, in check
    latest_url = get_latest(external_data.checker_data, "url-pattern", html)
  File "/home/zlopez/git/flatpak-external-data-checker/src/checkers/htmlchecker.py", line 47, in get_latest
    f"{pattern_name} {pattern} does not contain exactly 1 match group"
ValueError: url-pattern re.compile('https://download.gnome.org/sources/gnome-chess/(d+.d+)/gnome-chess-(d+.d+.d+).tar.xz') does not contain exactly 1 match group
```

As I understand there is issue in url-pattern, because the url for gnome apps is `https://download.gnome.org/sources/gnome-chess/3.36/gnome-chess-3.36.1.tar.xz`, so the version is there twice, but one time without patch number and in the other with patch number. I'm not sure how to adjust the manifest to make it work. Could you give me some advice?

wjt:
Remove the first set of parentheses:

```
https://download.gnome.org/sources/gnome-chess/\d+.\d+/gnome-chess-(\d+.\d+.\d+).tar.xz
```

We don't need to capture the first one. The second one is the full version number.

Zlopez:
I tried the change locally and now I got rid of the error, but got:
```
+ 2020-09-23 15:27:51,737 WARNING src.checkers.htmlchecker: gnome-chess-3.36.1.tar.xz had no available version information
+ 2020-09-23 15:27:51,737 WARNING src.checkers.htmlchecker: gnome-chess-3.36.1.tar.xz had no available URL
```
But I know the `3.38.0` is out. Is there something else I have wrong in manifest?

[...]

PR #92, opened 2020-10-08
Adds GNOME Checker

A6GibKm:
Adds a checker for gnome, see #91. Format should be
```json
                    "x-checker-data": {
                        "type": "gnome",
                        "name": "baobab"
                    }
```
TODO:

- [x] Update README
- [ ] Unit test for unstable package
- [x] Unit test for package already using latest
- [x] Use an example for the test which won't be updated, unlike libhandy
- [x] Skip unstable releases, e.g. `3.37.x` and `40.alpha`
- [x] Set `skip-unstable` by default
- ~~Fix CI~~ See #90

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,12 +214,20 @@ Both the version number and the download URL will be gathered from  [download.gn
 "x-checker-data": {
     "type": "gnome",
     "name": "PACKAGE-NAME",
-    "skip-unstable": true
 }
 ```
 
-where `PACKAGE-NAME` is set according to the name in [download.gnome.org/sources](https://download.gnome.org/sources).
-`skip-unstable` is optional and when set to `true` it will ignore unstable releases, e.g. "3.37.1" or "40.rc.2".
+where `PACKAGE-NAME` is set according to the package name as seen in [download.gnome.org/sources](https://download.gnome.org/sources).
+
+By default the checker skips unstable releases, e.g. releases tagged as "3.37.1" or "40.rc.2". To disable this behavior use
+
+```json
+"x-checker-data": {
+    "type": "gnome",
+    "name": "PACKAGE-NAME",
+    "skip-unstable": false
+}
+```
 
 ## License and Copyright
 
--- a/README.md
+++ b/README.md
@@ -206,6 +206,20 @@ for [Snapcraft](https://snapcraft.io/) packages:
 }
 ```
 
+#### GNOME checker
+
+Both the version number and the download URL will be gathered from  [download.gnome.org](https://download.gnome.org/). The format is:
+
+```json
+"x-checker-data": {
+    "type": "gnome",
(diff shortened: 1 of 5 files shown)


src/checkers/gnomechecker.py
```diff
@@ -0,0 +1,120 @@
...
+
+    if pattern.groups != 2:
+        raise ValueError(
+            f"{pattern} does not contain exactly 2 match group"
+        )
+
+    resp = urllib.request.urlopen(url)
+    html = resp.read().decode()
+
+    m = pattern.findall(html)
```
gasinvein: Why use regexps to parse json?

tests/test_gnomechecker.py
```diff
@@ -0,0 +1,90 @@
...
+        self.assertEqual(data.filename, "libhandy-0.81.0.tar.xz")
+        self.assertIsNotNone(data.new_version)
+        self.assertEqual(
+            data.new_version.version, "1.0.0",
+        )
+        self.assertIsInstance(data.new_version.size, int)
+        self.assertGreater(data.new_version.size, 0)
+        self.assertIsNotNone(data.new_version.checksum)
+        self.assertIsInstance(data.new_version.checksum, str)
+        print(data.new_version.url)
```
bilelmoussaoui:
```suggestion
```

A6GibKm: Woops.

src/checkers/gnomechecker.py
A6GibKm: I have to admit that I did not consider that. This seems simpler, I don't know much complexity it adds versus checking the regex only on each release of the json.

README.md
```diff
@@ -206,6 +206,20 @@ for [Snapcraft](https://snapcraft.io/) packages:
...
+```json
+"x-checker-data": {
+    "type": "gnome",
+    "name": "PACKAGE-NAME",
+    "skip-unstable": true
+}
+```
+
+where `PACKAGE-NAME` is set according to the name in [download.gnome.org/sources](https://download.gnome.org/sources).
+`skip-unstable` is optional and when set to `true` it will ignore unstable releases, e.g. "3.37.1" or "40.rc.2".
```
bilelmoussaoui: This should be the default behaviour.

A6GibKm: Many libraries and packages won't follow the 3.x.y scheme. But ok. I will set it as the default and false should be used for those cases.

src/checkers/gnomechecker.py
A6GibKm: Done.

A6GibKm:
I don't know of an example of a package for which its latests version is an unstable one and won't be updated any time soon.

gasinvein:
Superseded by #141

PR #93, opened 2020-10-09
Add Rust checker

tilosp:
Used by https://github.com/flathub/org.freedesktop.Sdk.Extension.rust-nightly/pull/9

Depends on https://github.com/flathub/actions/pull/8

### DIFF
--- a/src/checkers/rustchecker.py
+++ b/src/checkers/rustchecker.py
@@ -1,0 +1,43 @@
+import datetime
+import toml
+import logging
+import urllib.request
+
+from src.lib import utils
+from src.lib.externaldata import ExternalFile, Checker
+
+log = logging.getLogger(__name__)
+
+
+class RustChecker(Checker):
+    @staticmethod
+    def _should_check(external_data):
+        return external_data.checker_data.get("type") == "rust"
+
+    def check(self, external_data):
+        if not self._should_check(external_data):
+            log.debug("%s is not a rust type ext data", external_data.filename)
+            return
+
+        channel = external_data.checker_data.get("channel", "stable")
+        package_name = external_data.checker_data["package"]
+        target_name = external_data.checker_data["target"]
+
+        url = f"https://static.rust-lang.org/dist/channel-rust-{channel}.toml"
+
+        log.debug("Getting extra data info from %s; may take a while", url)
+        with urllib.request.urlopen(url) as response:
+            data = toml.loads(response.read().decode())
+
+        package = data["pkg"][package_name]
+        target = package["target"][target_name]
+        if target["available"]:
+            new_version = ExternalFile(
+                target["xz_url"],
+                target["xz_hash"],
+                None,
+                package["version"].split(" ")[0],
+                datetime.datetime.strptime(data["date"], "%Y-%m-%d"),
+            )
+            if not external_data.current_version.matches(new_version):
+                external_data.new_version = new_version
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,7 @@ RUN apt-get update \
       python3-ruamel.yaml \
       python3-setuptools \
       python3-tenacity \
+      python3-toml \
   && apt-get clean \
   && rmdir /var/cache/apt/archives/partial
 
--- a/README.md
+++ b/README.md
@@ -222,6 +222,20 @@ for [Snapcraft](https://snapcraft.io/) packages:
 }
 ```
 
+#### Rust checker
+
+Special checker that will check for available updates
+for [Rust](https://www.rust-lang.org/):
+
+```json
+"x-checker-data": {
+    "type": "rust",
+    "package": "package name, for example: rust",
+    "channel": "nightly, stable or beta",
+    "target": "target triple, for example: x86_64-unknown-linux-gnu"
+}
+```
+
 
 ## License and Copyright
 
--- a/src/checkers/__init__.py
+++ b/src/checkers/__init__.py
@@ -6,6 +6,7 @@ from .htmlchecker import HTMLChecker
 from .jetbrainschecker import JetBrainsChecker
 from .snapcraftchecker import SnapcraftChecker
 from .anityachecker import AnityaChecker
+from .rustchecker import RustChecker
 
 
 # For each ExternalData, checkers are run in the order listed here, stopping once data.state is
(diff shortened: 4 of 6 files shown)


tilosp:
Test failure is unrelated, see: https://github.com/flathub/flatpak-external-data-checker/pull/90#issuecomment-695816437

src/checkers/rustchecker.py
```diff
@@ -0,0 +1,37 @@
...
+    @staticmethod
+    def _should_check(external_data):
+        return external_data.checker_data.get("type") == "rust"
+
+    def check(self, external_data):
+        if not self._should_check(external_data):
+            log.debug("%s is not a rust type ext data", external_data.filename)
+            return
+
+        channel = external_data.checker_data["channel"]
```
bilelmoussaoui: I would default to stable here

bilelmoussaoui:
Do you mind rebasing the PR? :) Thanks

src/checkers/rustchecker.py
```diff
@@ -0,0 +1,37 @@
...
+        url = f"https://static.rust-lang.org/dist/channel-rust-{channel}.toml"
+
+        log.debug("Getting extra data info from %s; may take a while", url)
+        with urllib.request.urlopen(url) as response:
+            data = toml.loads(response.read().decode())
+
+        package = data["pkg"][package_name]
+        target = package["target"][target_name]
+        if target["available"]:
+            new_version = utils.get_extra_data_info_from_head(target["xz_url"])
```
gasinvein: Is there any reason to download the archive and not just use provided `xz_url` and `xz_hash` as is?

tilosp: `xz_hash` is actually used, see the line below this one

I have done it like this because it is the same way here:
https://github.com/flathub/flatpak-external-data-checker/blob/69ea8b5790853c0069f0a5ed57f13fb62fddb3a6/src/checkers/firefoxchecker.py#L167-L168

I think it's there to get the size of the file, not sure if that’s optional here.
But in any case the head request shouldn't download the file

tilosp:
rebased onto master

src/checkers/rustchecker.py
```diff
@@ -0,0 +1,39 @@
...
+            new_version = new_version._replace(
+                checksum=target["xz_hash"], version=package["version"]
```
gasinvein: `version` here looks something like `1.48.0 (7eac88abb 2020-11-16)`. I believe it is invalid for appstream. Maybe parse it with regex or else, and also extract date from it?

src/checkers/rustchecker.py
gasinvein: Size is useless for archive type sources anyway. I guess we shouldn't add it for anything but `extra-data`, but that's a separate issue.

src/checkers/rustchecker.py
tilosp: is there any documentation on what version numbers are valid?

src/checkers/rustchecker.py
tilosp: I am not sure how to construct a `ExternalFile` without a size.
Should I use some fixed bogus value to avoid the inexpensive `HEAD` request?

src/checkers/rustchecker.py
gasinvein: I don't know any appstream documentation other than [this](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-releases), and it doesn't seem to specify any constraints for version numbers, so maybe I'm wrong and it allows any string as a version. Still we may want to parse it in order to get upstream release date.

src/checkers/rustchecker.py
gasinvein: Maybe not. We could rewrite this later once we get rid of spurious `size` property.

src/checkers/rustchecker.py
tilosp: that’s not actually the release date, it is the commit date of the commit used for the release.
For example rust `1.48.0` was released on `2020-11-19` but the version looks like this `1.48.0 (7eac88abb 2020-11-16)`
See: https://github.com/rust-lang/rust/releases/tag/1.48.0

barthalion: I think it's fine to strip the extra info like commit or build date.

src/checkers/rustchecker.py
gasinvein: Now you can construct `ExternalFile`  manually and pass `None` as `size` value (after rebasing onto master).

src/checkers/rustchecker.py
tilosp: I agree that its fine for stable releases but different nightlies would then look the same for example `0.51.0-nightly`.

So we can either stick to the version strings that upstream provides or strip the extra info only on beta and stable.

gasinvein: >different nightlies would then look the same for example `0.51.0-nightly`

I'd say it's fine. Unlikely that someone would care about exact commit.

src/checkers/rustchecker.py
tilosp: done

gasinvein:
@tilosp Don't mind adding a test case for it? I guess you could reuse stripped down rust-nightly manifest as a test sample. Other than this, LGTM.

src/checkers/rustchecker.py
tilosp: added code to remove everything after and including the first `(`

src/checkers/rustchecker.py
```diff
@@ -0,0 +1,44 @@
...
+            data = toml.loads(response.read().decode())
+
+        package = data["pkg"][package_name]
+        target = package["target"][target_name]
+        if target["available"]:
+            new_version = ExternalFile(
+                target["xz_url"],
+                target["xz_hash"],
+                None,
+                re.match('[^(]+', package["version"]).group().strip(),
```
gasinvein: If you don't use regex to extract exact values, maybe it's easier to just split by whitespace? E.g. something like `package["version"].split(" ")[0]`

tilosp: done

tilosp:
> @tilosp Don't mind adding a test case for it? I guess you could reuse stripped down rust-nightly manifest as a test sample. Other than this, LGTM.

@gasinvein done I think this is ready to get merged now

gasinvein:
Seems fine. Thanks!

PR #94, opened 2020-10-29
Migrate tests to httpbingo.org

gasinvein:
httpbin seems abandoned, see postmanlabs/httpbin#617


gasinvein:
Now `black --check` fails on old files. @wjt should we allow it to reformat?

wjt:
Yeah, I think so – strange, I wonder what's changed, Black or the code?

gasinvein:
I didn't change anything here but httpbin domain, so it must be black.

wjt:
Thanks!

PR #95, opened 2020-12-14
Add anitya (release-monitoring) checker

### LINKED ISSUE #16: Use release-monitoring to detect latest version
For example, using [metadata from release-monitor.org](https://release-monitoring.org/project/17854/) for https://github.com/flathub/com.adobe.Flash-Player-Projector could allow detecting newer version.

However, Anitya/release-monitoring doesn't seem to record the URLs, and probably expects manifests to have placeholders for those.

It would still be useful for the Flatpak mentioned above because I could know about the latest version of the package even if the download URL didn't change, and it could modify the appdata file to insert the new version with the corresponding release date.

gasinvein:
A dead simple integration with https://release-monitoring.org (or any other hosted Anitya instance).
Given numeric `project-id`, retrieves latest version known to Anitya instance and substitutes it into given `url-template` to get download url.
Also covers "Github checker" use-case, since Anitya has Gihub support.
Closes #16

### DIFF
--- a/src/checkers/__init__.py
+++ b/src/checkers/__init__.py
@@ -5,6 +5,7 @@ from .urlchecker import URLChecker
 from .htmlchecker import HTMLChecker
 from .jetbrainschecker import JetBrainsChecker
 from .snapcraftchecker import SnapcraftChecker
+from .anityachecker import AnityaChecker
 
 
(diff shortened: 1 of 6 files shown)


gasinvein:
@wjt Do we need a test case for this? I believe there is virtually nothing to go wrong on our side, since we just download a json and take a single value from it.

PR #98, opened 2020-12-24
Rust: Use incrementing versions in appstream

gasinvein:
Should fix both flathub/org.freedesktop.Sdk.Extension.rust-nightly#12 and flathub/org.freedesktop.Sdk.Extension.rust-nightly#14
Will set versions in appstream to likes `1.48.0` for stable and e.g. `1.50.0-nightly-20201224` for nightly, based on release date.
@tilosp @bilelmoussaoui Does this look valid to you?

### DIFF
--- a/src/checkers/rustchecker.py
+++ b/src/checkers/rustchecker.py
@@ -1,12 +1,17 @@
 import datetime
-import toml
 import logging
 import urllib.request
+import re
+
+import toml
 
 from src.lib import utils
 from src.lib.externaldata import ExternalFile, Checker
 
 log = logging.getLogger(__name__)
+
+
+VERSION_RE = re.compile(r"^(\S+)\s+\((\S+)\s+(\S+)\)")
 
 
 class RustChecker(Checker):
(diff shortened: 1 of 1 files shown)


src/checkers/rustchecker.py
```diff
@@ -31,13 +36,21 @@ def check(self, external_data):
 
         package = data["pkg"][package_name]
         target = package["target"][target_name]
+
+        release_date = datetime.date.fromisoformat(data["date"])
+        version, _, _ = VERSION_RE.match(package["version"]).groups()
+        if channel == "stable":
```
tilosp: this case should probably include the `beta` channel
the versions look like this `1.49.0-beta.5 (b0dc3c6f4 2020-12-22)` so using `1.49.0-beta.5` should be fine

gasinvein: Exactly `beta` or any non-`nightly`?

tilosp: any non nightly seems best to me

tilosp (approved):
LGTM

bilelmoussaoui (approved):

PR #103, opened 2021-01-19
Rework tests action and docker image

gasinvein:
Make docker image usable for local runs or other external usage.
This needs to slightly alter run-tests workflow, installing dependencies for tests only as needed.


gasinvein:
@wjt @andrunko Is it ok to run as root inside container by default on EndlessOS? I can't get `bwrap` to work in rootless podman on Fedora with non-root user in container namespace.

Issue #33, opened 2019-08-26
Support updating to latest Git tag matching some pattern


wjt:
Vim's development model is that each commit to the master branch gets a tag, and this is the only way to get fixes since the first release on the stable branch. As a result, the Flatpak builds from a Git tag.

https://github.com/flathub/org.vim.Vim/blob/master/auto-update.py implements logic to update to the HEAD of master and open a PR. It could be combined with this checker.

However – the current cronjob for that script only runs it weekly, to avoid update churn for users and PR spam for me. Perhaps the checker could have some heuristic to reduce the number of PRs generated.

nanonyme:
We more or less have Git tracking available at https://gitlab.com/BuildStream/bst-external/blob/master/bst_external/sources/git_tag.py#L350-387

I was talking a bit about this on IRC, I think the same logic should be replicated here. Our approach supports majority of opensource projects using Git. I can take a look at this later if I have time. This is a non-trivial task if this should be implemented in a way that supports eg Gnome projects.

bilelmoussaoui:
> We more or less have Git tracking available at https://gitlab.com/BuildStream/bst-external/blob/master/bst_external/sources/git_tag.py#L350-387
> 
> I was talking a bit about this on IRC, I think the same logic should be replicated here. Our approach supports majority of opensource projects using Git. I can take a look at this later if I have time. This is a non-trivial task if this should be implemented in a way that supports eg Gnome projects.

I think for gnome projects we should provide a different checker that would use download.gnome.org instead and properly handle the stable/unstable releases.

nanonyme:
@bilelmoussaoui git_tag include/exclude can be used to configure stable/unstable updates like Gnome does.

nanonyme:
But sure, for Gnome one usually uses tarballs still.

gasinvein:
Work in progress, review/testing welcome #104

gasinvein:
#115 implements this feature.

PR #105, opened 2021-01-23
Add git checker

gasinvein:
This is similar to URL checker - it applies to all git sources and checks if the specified tag/branch is still valid.

Follow-up for #104, see also #33

### DIFF
--- a/src/checkers/__init__.py
+++ b/src/checkers/__init__.py
@@ -7,6 +7,7 @@ from .jetbrainschecker import JetBrainsChecker
 from .snapcraftchecker import SnapcraftChecker
 from .anityachecker import AnityaChecker
 from .rustchecker import RustChecker
+from .gitchecker import GitChecker
 
 
 # For each ExternalData, checkers are run in the order listed here, stopping once data.state is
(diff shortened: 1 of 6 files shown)


gasinvein:
@grulja @travier @tsdgeos  Hey there. I've seen some KDE apps on Flathub who set git tag as a branch property, like `"branch": "v1.0.0"`. Is it intentional, e.g. some convention for KDE apps?
flatpak-external-data-checker with this changes will submit PRs replacing `banch` with `tag` property (with same value). If it's not desirable - I need your advice how to handle it better.

tsdgeos:
Can you point to some of those?

gasinvein:
@tsdgeos Oddly enough, I can't find any such apps under org.kde. namespace anymore. Oh well, I'll then just assume that it's ok to replace branch with tag property where applicable. Sorry for bothering you.

gasinvein:
I've removed all the guesswork. If something other than actual branch is specified in a `branch` property, the checker won't be able to retrieve remote commit and, thus, won't try to update it.

PR #110, opened 2021-02-03
Fix jenkins integration

andrunko:
This fixes Endless Jenkins integration that got broken after 064fdeaa85e00bf72832d1991a571b4e009137ae.

~~It should still allow running the container as it is now but with some small caveats (e.g. if passing only a manifest as param to "docker run" it has to end with ".json", ".yml" or ".yaml").~~

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,6 +24,14 @@ RUN apt-get update && \
   && apt-get clean \
   && rmdir /var/cache/apt/archives/partial
 
+# Creating the user is required because jenkins runs he container
+# with the same user as the host (with '-u <uid>:<gid>')
+# but without the user existing 'git' fails with 'No user exists for uid ...'.
+ARG USER_ID=1000
+ARG GROUP_ID=1000
+RUN groupadd -g $GROUP_ID user && \
+    useradd -u $USER_ID -s /bin/sh -m -g user user
+
 COPY src /app/src
 COPY flatpak-external-data-checker /app/
 COPY canonicalize-manifest /app/
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -20,7 +20,7 @@ node('flatpak-builder') {
              * a workaround to make sure the user running the container exists
              * inside of it (see Dockerfile). This is needed because jenkins runs
              * the container with the same user as the host (with '-u <uid>:<gid>')
-             * but 'git push' fails with 'No user exists for uid ...' if the user
+             * but 'git' fails with 'No user exists for uid ...' if the user
              * doesn't exist inside the container (other option is to mount
              * '/etc/passwd' as a volume and export HOME/USER envs accordingly
              * when running the container ('image.inside' below)).
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -35,7 +35,7 @@ node('flatpak-builder') {
             try {
                 /* The '--privileged' param is required to run 'bwrap' which is used
                  * by the flatpak external data checker */
-                image.inside('--privileged') {
+                image.inside('--privileged --entrypoint=""') {
                     sshagent(credentials: [ 'fe45ca53-7c92-47db-b3b1-b8d0cc8507ed' ]) {
                         withCredentials([string(credentialsId: 'github-api-token-rw-jobs', variable: 'GITHUB_TOKEN')]) {
                             timeout(time: 20, unit: 'MINUTES') {


gasinvein:
Oh, did I break CI for Endless? So sorry. I was pretty sure the image was here only for user convenience.

Dockerfile
```diff
@@ -24,10 +24,20 @@ RUN apt-get update && \
...
+
+COPY docker-entrypoint.sh /
 COPY src /app/src
 COPY flatpak-external-data-checker /app/
 COPY canonicalize-manifest /app/
 
 RUN python3 -m compileall /app/src
 
-ENTRYPOINT [ "/app/flatpak-external-data-checker" ]
+ENTRYPOINT [ "/docker-entrypoint.sh" ]
```
gasinvein: Do you control docker cli args in Jenkins? If so, maybe it's better to pass `--entrypoint whatever` to `docker` instead of guessing what we should run in a shell script?

andrunko: Not sure tbh, the error comes from (which runs `cat` to test the container):
```
$ docker run -t -d -u 1004:1004 --privileged -w /home/jenkins/workspace/check-flatpak-external-apps-pipeline -v /home/jenkins/workspace/check-flatpak-external-apps-pipeline:/home/jenkins/workspace/check-flatpak-external-apps-pipeline:rw,z -v /home/jenkins/workspace/check-flatpak-external-apps-pipeline@tmp:/home/jenkins/workspace/check-flatpak-external-apps-pipeline@tmp:rw,z -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** -e ******** flatpak-external-data-checker:latest cat
$ docker top 54624e2f05c7096d62a00b012b29e5ff3734393f613e885990d9c1c299799e9f -eo pid,comm
ERROR: The container started but didn't run the expected command. Please double check your ENTRYPOINT does execute the command passed as docker run argument, as required by official docker images (see https://github.com/docker-library/official-images#consistency for entrypoint consistency requirements).
```

I could try passing `--entrypoint` to `image.inside` on Jenkinsfile to see if that would work, but wonder if it may still be useful (and best practice) to have an entrypoint allowing running arbitrary commands (bash for example for debugging shell)?

gasinvein: >it may still be useful (and best practice) to have an entrypoint allowing running arbitrary commands

Since the first argument is likely to be a file with arbitrary name, such wrapper feels a bit fragile. But I'm not strictly against it, it may be useful indeed.

andrunko: Pushed a new branch `fix-jenkins-integration-new` overriding the entrypoint via Jenkinsfile and will verify if it works and update here accordingly.

andrunko: Ok, it works, updated changes.

gasinvein (approved):
Seems fine.

Dockerfile
```diff
@@ -24,6 +24,14 @@ RUN apt-get update && \
   && apt-get clean \
   && rmdir /var/cache/apt/archives/partial
 
+# Creating the user is required because jenkins runs he container
+# with the same user as the host (with '-u <uid>:<gid>')
+# but without the user existing 'git' fails with 'No user exists for uid ...'.
```
gasinvein, line 29: I wonder why `git` behaves like this - why does it even care about local system user presence?

andrunko: Good question, no idea but it does...

andrunko:
> Seems fine.

Thanks, going ahead and landing this to unbreak our CI :).

gasinvein:
Hmm, since this image is now available in a public registry, maybe create another one based on it just for Endless CI? You can use the `jenkins-check-flatpak-external-apps` script there as an entrypoint, too.

Dockerfile
andrunko: Looking at the description on https://issues.jenkins.io/browse/JENKINS-47026, it seems not only git fails, ssh also (which is probably why git fails).

andrunko:
> Hmm, since this image is now available in a public registry, maybe create another one based on it just for Endless CI? You can use the `jenkins-check-flatpak-external-apps` script there as an entrypoint, too.

Interesting, I wasn't aware the image was public now, may change that at some point but orthogonal to the issue here. Also we may move our Jenkins integration to somewhere internal as there is not much point in having it public.

gasinvein:
It's on GHCR: [`ghcr.io/flathub/flatpak-external-data-checker`](https://github.com/orgs/flathub/packages/container/package/flatpak-external-data-checker)
I've reused it for flathub's github action image, adding some files on top of it and using a different entrypoint, thus, thought you can do something similar for Endless.

PR #117, opened 2021-02-08
Don't add new commit property to git sources

gasinvein:
When checking if git source is still valid, only check if the tag/branch tip still points to specified commit. If no commit was specified, don't add it.

I'm not entirely sure if we should add commit or not. Specifying commit is good for reproducibility, but the maintainer may not want it (e.g. if the tag is intentionally being re-tagged periodically).


andrunko (approved):
Makes sense and looks good to me.

PR #120, opened 2021-02-11
Add checker for PyPI packages

### LINKED ISSUE #74: Support updating PyPI modules
Python components which come from https://pypi.org/ should be easy enough to support. There are two cases:

- Apps which specify a single package from PyPI – https://github.com/flathub/org.learningequality.Kolibri is one example.
- Apps which generate a flattened list of packages from a dependency tree using https://github.com/flatpak/flatpak-builder-tools/tree/master/pip

gasinvein:
Given PyPI package name and desired dist type (`sdist` or `bdist_wheel`), this checker finds latest available release.
Fixes #74, see test manifest for usage example.
I would appreciate a review from @nanonyme who is more familiar with PyPI and previously worked on something similar for BuildStream.

### DIFF
--- a/src/checkers/pypichecker.py
+++ b/src/checkers/pypichecker.py
@@ -71,8 +71,13 @@ class PyPIChecker(Checker):
             response.raise_for_status()
             pypi_data = response.json()
 
+        if constraints:
+            releases = pypi_data["releases"]
+        else:
+            releases = {pypi_data["info"]["version"]: pypi_data["urls"]}
+
         downloads = sorted(
-            _filter_downloads(pypi_data["releases"], constraints, package_type),
+            _filter_downloads(releases, constraints, package_type),
             key=lambda r: r[2],
         )
 
--- a/src/checkers/pypichecker.py
+++ b/src/checkers/pypichecker.py
@@ -1,0 +1,92 @@
+import logging
+from datetime import datetime
+from distutils.version import StrictVersion, LooseVersion
+import operator
+import re
+import typing as t
+
+import requests
+
+from ..lib.externaldata import Checker, ExternalFile
+
+log = logging.getLogger(__name__)
+
+PYPI_INDEX = "https://pypi.org/pypi"
+OPERATORS = {
+    "<": operator.lt,
+    "<=": operator.le,
+    ">": operator.gt,
+    ">=": operator.ge,
+    "==": operator.eq,
+    "!=": operator.ne,
+}
+BDIST_RE = re.compile(r"^(\S+)-(\d[\d\.\w]*\d)-(\S+)-(\S+)-(\S+).whl$")
+
+
+def _version_matches(version: str, constraints: t.List[t.Tuple[str, str]]):
+    if not constraints:
+        return True
+    for ver_oper, ver_limit in constraints:
+        oper = OPERATORS[ver_oper]
+        try:
+            matches = oper(StrictVersion(version), StrictVersion(ver_limit))
+        except ValueError:
+            matches = oper(LooseVersion(version), LooseVersion(ver_limit))
+        return matches
+    return None
+
+
+def _filter_downloads(
+    pypy_releases: t.Dict[str, t.List[t.Dict]],
+    constraints: t.List[t.Tuple[str, str]],
+    packagetype: str,
+) -> t.Generator[t.Tuple[str, t.Dict, datetime], None, None]:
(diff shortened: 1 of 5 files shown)


nanonyme:
@gasinvein is the intent to allow wheels which contain compiled binaries?

src/checkers/pypichecker.py
```diff
@@ -0,0 +1,92 @@
...
+    constraints: t.List[t.Tuple[str, str]],
+    packagetype: str,
+) -> t.Generator[t.Tuple[str, t.Dict, datetime], None, None]:
+    for pypi_version, pypi_downloads in pypy_releases.items():
+        if not _version_matches(pypi_version, constraints):
+            continue
+        for download in pypi_downloads:
+            if download["packagetype"] != packagetype:
+                continue
+            if download["python_version"] not in ["source", "py3", "py2.py3"]:
```
nanonyme, line 50: Note that manylinux14 abi3 is not covered by this check. That is what is expected to be the future precompiled wheel format.

gasinvein:
> @gasinvein is the intent to allow wheels which contain compiled binaries?

In future maybe, but currently not. It should skip all wheel with ABI.

src/checkers/pypichecker.py
gasinvein: You mean such wheel will pass this check or not? The check current intent is to skip all wheels with native binaries.

nanonyme: Sounds like this is fine then.

nanonyme:
What about compatibility? There's no guarantee pip packages are actually compatible with each other if you just update one.

gasinvein:
I don't think there is much we can do about it, except to provide an option for limiting suitable package versions like this:
https://github.com/flathub/flatpak-external-data-checker/blob/ffbdbfd62984637c0154a618de531cc93a7e2d87/tests/com.valvesoftware.Steam.yml#L26-L31
(`==` here can be replaced with `<=`, `>` and alikes)
It's not documented yet because I didn't yet figure out a future-proof schema for version constraints in checker metadata.

src/checkers/pypichecker.py
```diff
@@ -0,0 +1,92 @@
...
+
+def _version_matches(version: str, constraints: t.List[t.Tuple[str, str]]):
+    if not constraints:
+        return True
+    for ver_oper, ver_limit in constraints:
+        oper = OPERATORS[ver_oper]
+        try:
+            matches = oper(StrictVersion(version), StrictVersion(ver_limit))
+        except ValueError:
+            matches = oper(LooseVersion(version), LooseVersion(ver_limit))
```
nanonyme, line 34: Iirc current version PEP is superset of this. It's probably close enough though.

src/checkers/pypichecker.py
```diff
@@ -0,0 +1,92 @@
...
+        if "versions" in external_data.checker_data:
+            constraints = [(o, l) for o, l in external_data.checker_data["versions"]]
+        else:
+            constraints = []
+
+        with self.session.get(f"{PYPI_INDEX}/{package_name}/json") as response:
+            response.raise_for_status()
+            pypi_data = response.json()
+
+        downloads = sorted(
```
nanonyme, line 79: Just pick the version advertised as latest if there are no constraints. Going over releases can be really slow for certain packages with massive history.

gasinvein: But what if there is no requested packagetype among the latest release downloads?

nanonyme: Maybe then just provide no update. That is most likely a mistake in upstream. Picking random version may just end in a mess.

gasinvein: Fair enough. Done.

nanonyme:
Looks good to me

gasinvein:
Hmm, AFAIK python allows only one version constraint in requirements. Maybe we should do the same? It will reduce complexity a bit.

nanonyme:
It actually supports multiple in same line so you can declare a range.

gasinvein:
Well then. Didn't know this, how it looks like?

nanonyme:
@gasinvein https://github.com/psf/requests/blob/master/setup.py#L45, basically you just have comma as separator.

gasinvein:
@nanonyme And the comma does conjunction, right? If so, I guess the checker version constraints would behave more or less same as requirement notation.

nanonyme:
Yes, indeed. Checker should work fine as is.

PR #123, opened 2021-02-12
ci: Push all tags

gasinvein:
The workflow stopped pushing all docker image tags to registry - looks like only `latest` is pushed. Try to fix this.


gasinvein:
Looks like Docker version was updated on runners. Now it supports `--all-tags` option.

PR #125, opened 2021-02-15
Use requests module to download files

gasinvein:
requests module is easier to use, has builtin debug logging and it provides a builtin mechanism for retries, so we can get rid of `tenacity` dependency (we already depend on `requests` through `PyGithub`).

This PR shouldn't change behavior (except for default User-Agent change), just some internal work. But since it's more of a personal preference, asking for @wjt and @andrunko consent.

### DIFF
--- a/src/lib/utils.py
+++ b/src/lib/utils.py
@@ -36,14 +36,8 @@ import typing as t
 
 from collections import OrderedDict
 from ruamel.yaml import YAML
-from tenacity import (
-    before_sleep_log,
-    retry,
-    retry_if_exception_type,
-    stop_after_attempt,
-    wait_fixed,
-)
 from elftools.elf.elffile import ELFFile
+import requests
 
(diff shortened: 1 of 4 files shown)


src/lib/utils.py
```diff
@@ -36,14 +36,8 @@
 
 from collections import OrderedDict
 from ruamel.yaml import YAML
-from tenacity import (
```
wjt, line 39: Can we drop a dependency on tenacity?

wjt (commented):
yes plz

src/lib/utils.py
gasinvein: You mean drop it from docker image?

wjt: Yes

gasinvein: Dropped.

wjt (approved):

Issue #128, opened 2021-02-17
XML parse error when metainfo starts with a comment?


gasinvein:
flatpak-external-data-checker can't update `com.microsoft.Edge` (and probably some others apps) due to exception in XML parsing:
```
Traceback (most recent call last):
  File "/usr/lib/python3.7/xml/sax/expatreader.py", line 217, in feed
    self._parser.Parse(data, isFinal)
xml.parsers.expat.ExpatError: XML or text declaration not at start of entity: line 2, column 0

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/app/flatpak-external-data-checker", line 30, in <module>
    main()
  File "/app/src/main.py", line 301, in main
    if run_with_args(parse_cli_args()) > 0:
  File "/app/src/main.py", line 283, in run_with_args
    changes = manifest_checker.update_manifests()
  File "/app/src/checker.py", line 286, in update_manifests
    self._update_manifest(path, datas, changes)
  File "/app/src/checker.py", line 277, in _update_manifest
    appdata, last_update.version, timestamp.strftime("%F")
  File "/app/src/lib/appdata.py", line 139, in add_release_to_file
    add_release(appdata_path, buf, version, date)
  File "/app/src/lib/appdata.py", line 134, in add_release
    reader.parse(in_path_or_stream)
  File "/usr/lib/python3.7/xml/sax/saxutils.py", line 312, in parse
    self._parent.parse(source)
  File "/usr/lib/python3.7/xml/sax/expatreader.py", line 111, in parse
    xmlreader.IncrementalParser.parse(self, source)
  File "/usr/lib/python3.7/xml/sax/xmlreader.py", line 125, in parse
    self.feed(buffer)
  File "/usr/lib/python3.7/xml/sax/expatreader.py", line 221, in feed
    self._err_handler.fatalError(exc)
  File "/usr/lib/python3.7/xml/sax/saxutils.py", line 246, in fatalError
    self._err_handler.fatalError(exception)
  File "/usr/lib/python3.7/xml/sax/handler.py", line 38, in fatalError
    raise exception
xml.sax._exceptions.SAXParseException: /github/workspace/flathub/com.microsoft.Edge/com.microsoft.Edge.metainfo.xml:2:0: XML or text declaration not at start of entity
```
I have no idea what's wrong. Comment on the first line of an XML file should be fine, shouldn't it?

TheEvilSkeleton:
[...]

PR #130, opened 2021-02-18
Collect and report problems during update

gasinvein:
As originally suggested in https://github.com/flathub/flatpak-external-data-checker/issues/128#issuecomment-780864107, if we got some non-fatal error during run, we should collect them and report to the maintainer.

I'm not sure if we should include errors in commit messages. It doesn't feel like the right place for error logs.

### DIFF
--- a/src/checker.py
+++ b/src/checker.py
@@ -73,6 +73,16 @@ def _external_source_filter(manifest_path: str, source) -> t.Optional[bool]:
     return True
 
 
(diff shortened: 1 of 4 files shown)


gasinvein:
Or maybe instead of inventing a new abstraction we can collect logging messages?

gasinvein:
This can be done better via exception handling.

PR #131, opened 2021-02-19
Improve stock messages

wjt:
- Stop telling people to contact individual humans
- Split the automerge message in 2
- Add emoji :robot: 

Danger danger: completely untested.


src/main.py
```diff
@@ -128,17 +128,23 @@ def commit_changes(changes):
...
 DISCLAIMER = (
-    "<i>(This pull request was automatically generated by "
-    "[flathub/flatpak-external-data-checker]"
+    "🤖 This pull request was automatically generated by "
+    "[flathub/flatpak-external-data-checker] "
     "(https://github.com/flathub/flatpak-external-data-checker). "
-    "Please contact or mention `@barthalion` or `@wjt` if you have any "
-    "questions or complaints.)</i>"
+    "Please [open an issue] "
+    "(https://github.com/flathub/flatpak-external-data-checker/issues/new) "
```
gasinvein: I guess it will result in broken markup. Put the hyperlink on single line just in case, or otherwise make sure there is no whitespace between `]` and `(`.

src/main.py
```diff
@@ -180,6 +186,7 @@ def open_pr(subject, body, branch, manifest_checker=None):
         repocfg = {}
 
     automerge = repocfg.get("automerge-flathubbot-prs")
+    has_broken_source = False
 
     # Enable automatic merge if extra-data is broken unless explicitly disabled
     if automerge is not False and manifest_checker:
```
gasinvein: This check is redundant if you're setting another variable (`automerge` and `has_broken_source` can both be `True` simultaneously).

src/main.py
```diff
@@ -128,17 +128,23 @@ def commit_changes(changes):
 
 
 DISCLAIMER = (
-    "<i>(This pull request was automatically generated by "
-    "[flathub/flatpak-external-data-checker]"
+    "🤖 This pull request was automatically generated by "
+    "[flathub/flatpak-external-data-checker] "
     "(https://github.com/flathub/flatpak-external-data-checker). "
```
gasinvein: Same here, probably spurious whitespace between `]` and `(`.

src/main.py
wjt: I've removed the whitespace but kept them on separate lines, as they were before.

src/main.py
gasinvein: Oh wait. Sorry. It's not redundant. It's checking that automerge wasn't explicitly disabled.
Still, since the "should we force automerge" value now sits in another variable, maybe move the check somewhere else, rename the variable to something like `force_automerge`, or keep changing `automerge` value instead of a new variable as it was before?

wjt: Keeping them separate allows us to use different messages in each case. I've rephrased this loop and renamed the variable.

gasinvein (approved):
LGTM, but still it would be good to test this somehow.

wjt:
It would… I had a quick look at using unittest.mock but don't really have time to sink into it. I can just wait till a repo I use this on (e.g. dropbox) has a PR and test it by hand.

gasinvein:
After looking through the changes once more, I think the logic stays the same, so probably it's safe to merge.
We can look into autotesting `main.py` later.

PR #141, opened 2021-02-24
Add GNOME checker v2

gasinvein:
Retrieve latest version tarball from download.gnome.org for given project name.
Supersedes #92

### DIFF
--- a/src/checkers/gnomechecker.py
+++ b/src/checkers/gnomechecker.py
@@ -1,0 +1,80 @@
+import logging
+from urllib.parse import urljoin
+import typing as t
+
+import requests
+
+from ..lib.externaldata import Checker, ExternalData, ExternalFile
+
+log = logging.getLogger(__name__)
+
+GNOME_MIRROR = "https://download.gnome.org/"
+
+
+def _parse_checksums(text: str) -> t.Dict[str, str]:
+    result = {}
+    for line in text.splitlines():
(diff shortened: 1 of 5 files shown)


src/checkers/gnomechecker.py
```diff
@@ -0,0 +1,80 @@
...
+        digest, filename = line.strip().split(maxsplit=1)
+        result[filename] = digest
+    return result
+
+
+def _is_stable(version: str) -> bool:
+    major, minor = version.split(".")[:2]
+    if int(major) >= 40:
+        return minor not in ["alpha", "beta", "rc"]
+    return (int(minor) % 2) == 0
```
gasinvein, line 26: @bilelmoussaoui  I'm not sure about the algorithm to check if GNOME project version is stable. Is it ok to assume that all versions prior to 40 had minor version number being odd for pre-releases?

tests/test_gnomechecker.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python3
+# Copyright © 2020 Maximiliano Sandoval <msandova@protonmail.com>
```
wjt:
```suggestion
# Copyright © 2020–2021 Maximiliano Sandoval <msandova@protonmail.com>
```

gasinvein:
Well, I guess we can fix stable versions detection algorithm later if it proves to be incorrect.

PR #145, opened 2021-03-02
Dockerfile: Install ssh-client

andrunko:
Commit b700d97a85 changed the Docker image to stop installing apt recommends and with that the image has now stopped pulling ssh-client which broke auth when using git+ssh (as used by the Jenkinsfile/Endless CI integration).

Let's re-add ssh-client manually.

PR #147, opened 2021-03-03
Exit with success if PR can be opened

wjt:
Previously, this script would exit(0) if some sources were out of date
but we were able to open a PR to update them. We want this behaviour
because the owner of whatever CI job runs the checker should not get a
failure email if the script opens a PR: that's a success for the
script! The script should only indicate failure if it is not able to
carry out the task it was asked to perform (edit the source tree, open a
PR, etc.).

f4cddf5d12579c39e1fac58f1730d5f553a4c58b changed this behaviour. Change
it back. (I missed this while reviewing that!)

@andrunko noticed this behaviour after #145 merged because the job successfully opened a PR for our Chrome Flatpak but then the Jenkins job still failed.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -294,7 +294,7 @@ def run_with_args(args):
         if args.update or args.commit_only or args.edit_only:
             changes = manifest_checker.update_manifests()
             if args.edit_only:
-                return 0
+                return (outdated_num, True)
             if changes:
                 with indir(os.path.dirname(args.manifest)):
                     subject, body, branch = commit_changes(changes)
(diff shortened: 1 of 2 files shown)


wjt:
I expect this will break the tests, though... :)

wjt:
It did indeed break the tests.

I'm not able to work further on this right now – one option could be to add a command-line parameter that enables the behaviour the tests expect.

gasinvein:
Oh. I understand I screwed up again, but don't fully understand how exactly. We did exit with status >0 if there were updates before f4cddf5, what changed after?

gasinvein:
Now I see. We should exit with status 0 if there was outdated data, but we successfully performed an update, i.e. opened a PR, committed changes or edited files, right?
If so, I suggest to explicitly check if we were asked to perform an update in `main()`.

gasinvein:
Or return a tuple from `run_with_args()`, like `(outdated_num: int, updated: bool)` and check it's value in `main()`. @wjt don't mind if I push to your branch?

wjt:
> Or return a tuple from `run_with_args()`, like `(outdated_num: int, updated: bool)` and check it's value in `main()`. @wjt don't mind if I push to your branch?

Sure, sounds good to me!

gasinvein:
Here, tests should be fine now. But please re-check me if it does what it's supposed to do.

gasinvein (approved):
Tested a bit on an outdated manifest, seems fine, with `--commit-only` returns 0, without args returns 1.

wjt:
That's great, thank you!

PR #151, opened 2021-03-04
htmlchecker: Don't follow redirects by default

### LINKED ISSUE #150: Sourceforge round-robin servers create flood of updates
The flathubbot that runs flatpak-external-data-checker created a flood of updates from version 6.5 to version 6.5 for Sweethome3D, while pointing to a slightly different tarball host each time.

These are the PRs created:
https://github.com/flathub/com.sweethome3d.Sweethome3d/pulls?q=is%3Apr+%226.5%22

And this is the `x-checker-data` field:
https://github.com/flathub/com.sweethome3d.Sweethome3d/blob/ee6fd8e4d491ebcbf11705b4b5b7589d0c9d3faf/com.sweethome3d.Sweethome3d.json#L38-L45

hadess:
This fixes using the HTML checker on sourceforge round-robin download
URLs with the redirect pointing to a different download server each
time.

Closes: #150

### DIFF
--- a/src/checkers/htmlchecker.py
+++ b/src/checkers/htmlchecker.py
@@ -63,12 +63,12 @@ class HTMLChecker(Checker):
         assert self.should_check(external_data)
 
         url = external_data.checker_data["url"]
-        combo_pattren = _get_pattern(external_data.checker_data, "pattern")
+        combo_pattern = _get_pattern(external_data.checker_data, "pattern")
         version_pattern = _get_pattern(external_data.checker_data, "version-pattern")
         url_pattern = _get_pattern(external_data.checker_data, "url-pattern")
         url_template = external_data.checker_data.get("url-template")
         sort_matches = external_data.checker_data.get("sort-matches", True)
-        assert combo_pattren or (version_pattern and (url_pattern or url_template))
+        assert combo_pattern or (version_pattern and (url_pattern or url_template))
 
         with requests.get(url) as response:
             response.raise_for_status()
@@ -77,11 +77,11 @@ class HTMLChecker(Checker):
         latest_version: t.Optional[str] = None
         latest_url: t.Optional[str] = None
 
-        if combo_pattren:
-            assert combo_pattren.groups == 2
+        if combo_pattern:
+            assert combo_pattern.groups == 2
             latest_pair = _get_latest(
                 html,
-                combo_pattren,
+                combo_pattern,
                 (lambda m: LooseVersion(m[1])) if sort_matches else None,
             )
             if latest_pair:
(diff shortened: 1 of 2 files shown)


wjt:
Seems fine if the tests pass!

gasinvein:
Maybe just always disable it? I can't imagine a case when a real (redirected) URL can be needed. All the other checkers that specify URL explicitly (all but URLChecker) use original URL.

hadess:
> Maybe just always disable it? I can't imagine a case when a real (redirected) URL can be needed. All the other checkers use original URL.

Trying that in the latest push. I'll update the PR description if we choose this implementation.

hadess:
How do I run those tests on my system? I wish it threw a diff as to what it would reformat instead of just saying "would reformat, Oh no!"

gasinvein:
@hadess Just run `black` against the source directory.

src/checkers/htmlchecker.py
```diff
@@ -109,11 +109,10 @@ def check(self, external_data):
 
         self._update_version(external_data, latest_version, abs_url)
 
-    def _update_version(
-        self, external_data, latest_version, latest_url, follow_redirects=True
-    ):
+    def _update_version(self, external_data, latest_version, latest_url):
```
gasinvein: This method is used by inherited checkers. Instead of removing the arg, change it's default value (all inherited checkers set `follow_redirects` to `False` explicitly anyway).

gasinvein:
The test fails exactly because the new URL received from HTMLChecker checking SourceForge is now different. I guess we should change this regex
https://github.com/flathub/flatpak-external-data-checker/blob/4ab679a71450774c7ea45d5125dbff7ae97a0cf9/tests/test_htmlchecker.py#L118
to something like
```
^https://sourceforge\.net/projects/qrupdate/.+/qrupdate-\d[\d\.]+\d\.tar\.gz$
```

tests/test_htmlchecker.py
```diff
@@ -115,7 +115,7 @@ def _test_combo_pattern_nosort(self, data):
         )
         self.assertRegex(
             data.new_version.url,
-            r"^https://[\d\w\.-]+sourceforge.net/project/qrupdate/qrupdate/[\d\.]+/qrupdate-[\d\.]+.tar.gz$",  # noqa: E501
+            r"^https://sourceforge\.net/projects/qrupdate/.+/qrupdate-\d[\d\.]+\d\.tar\.gz$", # noqa: E501
```
gasinvein: One whitespace before `#` was lost.

gasinvein (approved):
LGTM. @wjt Any objections?

PR #152, opened 2021-03-04
ci: Tell black to output a diff

wjt:
As suggested on #151.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -29,7 +29,7 @@ jobs:
 
       - name: Run black
         run: |
-          python3 -m black --check .
+          python3 -m black --check --diff .
 
   check-typing:
     name: Static type check


wjt:
So, that works.

But maybe it would be better to make it commit the change back to the repo, à la https://github.com/psf/Black/actions/runs/17913292/workflow

gasinvein:
> But maybe it would be better to make it commit the change back to the repo, à la https://github.com/psf/Black/actions/runs/17913292/workflow

I personally don't like auto commits, they add noise and make editing commits trickier.

wjt:
The intention would be that the submitter would pull & squash the commit into their patch(es), saving them having to copy-paste a diff into their source tree.

gasinvein:
Yeah, but one would need to keep that in mind and not forget to pull before adding more commits.

wjt:
@hadess re. your comment on https://github.com/flathub/flatpak-external-data-checker/pull/151#issuecomment-790499820, would you have preferred:

a) the CI job printed a diff showing what changes you should apply; or
b) the CI job pushed a fixup commit to your branch (if it could) that you could squash?

wjt:
Oh, well, it seems from reading https://github.com/cclauss/autoblack and https://github.community/t/cant-push-to-forked-repository-on-the-original-repositorys-pull-request-with-github-token/16287/2#M2372 that CI can't commit back to a fork, so I guess I'll drop that bit and just leave the change to output the diff.

hadess:
> a) the CI job printed a diff showing what changes you should apply; or

A diff would have been fine. The CI job with a fixup commit wouldn't know how to fix up PRs with multiple commits (unless it ran the job on each commit individually, but then you'll likely run into conflicts).

gasinvein (approved):
LGTM

Issue #154, opened 2021-03-11
Git checker doesn't get timestamps, resulting in unwanted PRs


wjt:
On a whim I wrote https://github.com/flathub/org.gnome.Boxes.Extension.OsinfoDb/pull/2 to automatically update the database used by GNOME Boxes when it is tagged upstream. When reviewing my own PR I noticed that the date added in the appdata is today's date, rather than the date attached to the tag.

It doesn't seem to be possible to ask `git ls-remote` to include more commit metadata. I guess the "easiest" thing to do would be to make a shallow clone of the repository at exactly the tag in question, then read the timestamp out of the tag; but it would be nice to avoid having to fetch the commit contents. I don't know Git well enough to know if there is an operation that can do this. (The GitHub API has a way! But that's not useful in cases, like this one, which aren't on GitHub.)

gasinvein:
Sure it would be a nice feature to have.
But I believe it won't make much difference in practice in most cases, because f-e-d-c on Flathub is likely to get the new tag about 1 hour after it was tagged, so the current date is likely to match the tag date anyway.

wjt:
The name of the branches that f-e-d-c pushes are based on the tree hash, which will change whenever the date changes, so if a PR is not merged within the same UTC day as being opened, it'd open another one because the hash would change.

Maybe theoretical? Maybe acceptable?

gasinvein:
Didn't think about it. Yeah, that would be not great.

gasinvein:
Maybe we should add GitHub and GitLab checkers instead, that would use API?
Or even add a special case for `git_ls_remote()` helper - if the repo is hosted on GitHub or GitLab, use their APIs to get commit timestamp.
I really would like to avoid cloning whole repos. This would make checking git sources significantly slower.

wjt:
That could work. The GitLab API for fetching a tag's metadata is unauthenticated if the repo is public:

```
$ http GET https://gitlab.gnome.org/api/v4/projects/gnome%2Fgnome-initial-setup/repository/tags/3.38.2
HTTP/1.1 200 OK
Cache-Control: max-age=0, private, must-revalidate
Connection: keep-alive
Content-Length: 1699
Content-Type: application/json
Date: Fri, 12 Mar 2021 08:45:04 GMT
Etag: W/"d7088e5a56582b820054cfdcea731b6d"
Server: nginx
Vary: Origin
X-Content-Type-Options: nosniff
[...]

PR #155, opened 2021-03-12
jsonchecker: Get and parse temestamp if provided

gasinvein:
Allows using timestamp from JSON APIs. `timestamp-query` must return date in ISO format, should work for GitHub and GitLab.


gasinvein:
@wjt Will this address your gitlab.gnome.org use-case as described in #154?

PR #160, opened 2021-03-26
Catch and retry partial read (v2)

### LINKED ISSUE #142: Incomplete HTTP read is ignored
In past week, flatpak-external-data-checker submitted several PRs (e.g. flathub/com.wps.Office#96) to WPS Office flatpak where the checksum (and only it) was changed. The "new" checksum is different each time and seems invalid - re-checking showed that the checksum didn't actually change.
It doesn't look like partial read, since the size doesn't change, and no http error occurs.

### LINKED ISSUE #159: Encoded content is decoded resulting in wrong size/checksum
If I understood it right, flatpak expects raw http data for `extra-data`, i.e. if it has `Content-Encoding`, it's not taken into account and size and checkum apply to undecoded content. Probably this behavior depends on libcurl/libsoup?
Our http client automatically decodes content if it detects `Content-Encoding`, so we calculate size/checksum for decoded content, which is wrong for flatpak.
On Flathub I've found only one app with `gzip`'ed upstream URL for extra-data - [GravitDesigner](https://github.com/flathub/io.designer.GravitDesigner).

gasinvein:
This is an another attempt of #143, but this time we make sure to return correct URL if there were redirects.
- urllib3 catches incomplete http read (fixes #142)
  - it can't retry incomplete read in streaming mode, so we preload small files to memory and stream only large ones
- explicitly disable content decoding to fix #159

### DIFF
--- a/src/checkers/htmlchecker.py
+++ b/src/checkers/htmlchecker.py
@@ -26,6 +26,7 @@ from distutils.version import LooseVersion
 import typing as t
 
 import requests
+import urllib3
 
 from ..lib import utils
 from ..lib.externaldata import ExternalData, Checker
@@ -120,9 +121,8 @@ class HTMLChecker(Checker):
                 latest_url, follow_redirects
             )
         except (
-            requests.exceptions.HTTPError,
-            requests.exceptions.ConnectionError,
-            requests.exceptions.ChunkedEncodingError,
+            urllib3.exceptions.HTTPError,
+            urllib3.exceptions.ConnectionError,
         ) as e:
             log.warning("%s returned %s", latest_url, e)
             external_data.state = ExternalData.State.BROKEN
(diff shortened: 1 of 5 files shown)


wjt:
I haven't reviewed the changes in detail but some of the logic is quite obscure – I think it would be worth documenting in comments in the code why we have this threshold for streaming versus reading into memory, etc.

gasinvein:
In the meantime I've started a global asyncio-based rewrite of some parts of f-e-d-c (to use non-blocking network IO and significantly fasten checks). Since it'll completely replace `requests`/`urllib3` with `aiohttp`, I'm not sure if this PR makes sense anymore.

gasinvein:
Closing in favor of #162, which supersedes changes made here.

PR #162, opened 2021-04-01
Run checks concurrently

### LINKED ISSUE #132: Some YAML anchors are removed on update
As discovered in flathub/com.valvesoftware.Steam.CompatibilityTool.Proton-Exp#5, it appears that YAML anchors sitting on the same line as their values get removed and replaced by their values on yaml dump.

Example:
```patch
-  cflags: &optflags -O2 -pipe -march=nocona -mtune=core-avx2 -mfpmath=sse -fwrapv -fno-strict-aliasing
+  cflags: -O2 -pipe -march=nocona -mtune=core-avx2 -mfpmath=sse -fwrapv -fno-strict-aliasing
   cflags-override: true
-  cxxflags: *optflags
+  cxxflags: -O2 -pipe -march=nocona -mtune=core-avx2 -mfpmath=sse -fwrapv -fno-strict-aliasing
   cxxflags-override: true
```

While this doesn't change anything for flatpak-builder, the maintainer has used anchors for a reason, so we should try harder to preserve them.

Unconfirmed, but it seems like `ruamel.yaml` version 0.16 keeps one-line anchors as expected.

### LINKED ISSUE #142: Incomplete HTTP read is ignored

gasinvein:
f-e-d-c downloads a lot of data from remote sources. Doing this sequentially, waiting on each transfer, is rather slow.
This PR makes (almost) all network I/O non-blocking with help of `asyncio` and `aiohttp`, downloading files concurrently and drastically reducing check times for app manifests with multiple modules.
For example, Steam (25 data items) check time goes down from 17 to 3 seconds (on my machine).

Changes include:
- Update Debian image to bullseye (asyncio in python 3.5 is way too old)
- Add context manager to checker class
  - set up aiohttp session per checker instance
  - create checker instance right before applying it, one per data item
- Split out individual data check to separate method; these are ran concurrently
- Make `git ls-remote` calls async, so these are not blocking, too
- Do new upstream version checks (HTTP API calls and such) in checkers with aiohttp, too

Using aiohttp fixes #142. Using newer ruamel.yaml seems to fix #132.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:buster
+FROM debian:bullseye
 
 ENV DEBIAN_FRONTEND=noninteractive
 
@@ -10,6 +10,7 @@ RUN apt-get update && \
       gir1.2-json-1.0 \
       python3-apt \
       python3-defusedxml \
+      python3-aiohttp \
       python3-gi \
       python3-cairo \
       python3-requests \
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ PyGithub
 ruamel.yaml
 toml
 pyelftools
+aiohttp
--- a/src/checker.py
+++ b/src/checker.py
@@ -21,6 +21,8 @@
 from collections import OrderedDict
 import datetime
 import typing as t
+import asyncio
+from dataclasses import dataclass
 
 from .checkers import ALL_CHECKERS
 from .lib.appdata import add_release_to_file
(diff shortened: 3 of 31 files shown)


gasinvein:
The large async-rewrite commit was split into two to ease review: the first one (156dbb2) adds context manager and wraps several methods with asyncio (while keeping actual IO synchronous and check() code unchanged), and the second one (0d92bdd) changes downloading methods to use aiohttp so that they actually run concurrently.

PR #165, opened 2021-04-10
htmlchecker and anityachecker: support templates using substrings of …

### LINKED ISSUE #134: Support building URLs and Git tags from $major_version, $minor_version and $patch_version
The download URLs of source archives of projects hosted on Sourceforge sometimes contain the pattern $major_version.$minor_version, without the $patch_version:

https://sourceforge.net/projects/qrupdate/files/qrupdate/1.2/qrupdate-1.1.2.tar.gz
https://downloads.sourceforge.net/project/flightgear/release-2020.3/flightgear-2020.3.6.tar.bz2

Thus, it would be nice if one could use these variables for the URL template.

Alexander-Wilms:
…version

Fixes #134

I'm new to Python, so I'm not sure if I missed something.

### DIFF
--- a/src/checkers/anityachecker.py
+++ b/src/checkers/anityachecker.py
@@ -1,7 +1,6 @@
 import logging
 import urllib.request
 import urllib.parse
-from string import Template
 
 import requests
 
@@ -40,7 +39,7 @@ class AnityaChecker(HTMLChecker):
 
     def _check_data(self, external_data, latest_version):
         url_template = external_data.checker_data["url-template"]
-        latest_url = Template(url_template).substitute(version=latest_version)
+        latest_url = self._substitute_placeholders(url_template, latest_version)
 
         self._update_version(
             external_data, latest_version, latest_url, follow_redirects=False
@@ -48,7 +47,7 @@ class AnityaChecker(HTMLChecker):
 
     def _check_git(self, external_data, latest_version):
         tag_template = external_data.checker_data["tag-template"]
-        latest_tag = Template(tag_template).substitute(version=latest_version)
+        latest_tag = self._substitute_placeholders(tag_template, latest_version)
 
         new_version = ExternalGitRef(
             url=external_data.current_version.url,
--- a/src/checkers/htmlchecker.py
+++ b/src/checkers/htmlchecker.py
@@ -92,7 +92,7 @@ class HTMLChecker(Checker):
                 html, version_pattern, LooseVersion if sort_matches else None
             )
             if latest_version and url_template:
-                latest_url = Template(url_template).substitute(version=latest_version)
+                latest_url = self._substitute_placeholders(url_template, latest_version)
             elif url_pattern:
                 assert url_pattern.groups == 1
                 latest_url = _get_latest(
@@ -108,6 +108,21 @@ class HTMLChecker(Checker):
         abs_url = urllib.parse.urljoin(base=url, url=latest_url)
 
         self._update_version(external_data, latest_version, abs_url)
+
+    @staticmethod
+    def _substitute_placeholders(template_string, version):
+        version_list = LooseVersion(version).version
+        tmpl = Template(template_string)
+        tmpl_vars = {"version": version}
+        for i, version_part in enumerate(version_list):
+            tmpl_vars[f"version{i}"] = version_part
+            if i == 0:
+                tmpl_vars["major"] = version_part
+            elif i == 1:
+                tmpl_vars["minor"] = version_part
+            elif i == 2:
+                tmpl_vars["patch"] = version_part
+        return tmpl.substitute(**tmpl_vars)
 
     def _update_version(
         self, external_data, latest_version, latest_url, follow_redirects=False
--- a/tests/org.flatpak.Flatpak.yml
+++ b/tests/org.flatpak.Flatpak.yml
@@ -21,3 +21,24 @@ modules:
           type: anitya
           project-id: 6377
           url-template: https://github.com/flatpak/flatpak/releases/download/$version/flatpak-$version.tar.xz
+          
+  - name: boost
+    sources:
+      - type: archive
+        url: https://dl.bintray.com/boostorg/release/1.74.0/source/boost_1_74_0.tar.bz2
+        sha256: 83bfc1507731a0906e387fc28b7ef5417d591429e51e788417fe9ff025e116b1
+        x-checker-data:
+          type: anitya
+          project-id: 6845
+          stable-only: true
+          url-template: https://dl.bintray.com/boostorg/release/$version/source/boost_${major}_${minor}_$patch.tar.bz2
+
+  - name: ghostscript
+    sources:
+      - type: archive
+        url: https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs926/ghostscript-9.26.tar.xz
+        sha256: 90ed475f37584f646e9ef829932b2525d5c6fc2e0147e8d611bc50aa0e718598
+        x-checker-data:
+          type: anitya
+          project-id: 1157
+          url-template: https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs$version0$version1$version2/ghostscript-$version.tar.xz
--- a/tests/test_anytiachecker.py
+++ b/tests/test_anytiachecker.py
@@ -16,9 +16,39 @@ class TestAnityaChecker(unittest.TestCase):
         checker = ManifestChecker(TEST_MANIFEST)
         ext_data = checker.check()
 
-        self.assertEqual(len(ext_data), 2)
+        self.assertEqual(len(ext_data), 4)
         for data in ext_data:
-            if data.filename == "flatpak-1.8.2.tar.xz":
+            if data.filename == "ghostscript-9.26.tar.xz":
+                self.assertIsNotNone(data.new_version)
+                self.assertIsInstance(data.new_version, ExternalFile)
+                self.assertRegex(
+                    data.new_version.url,
+                    r"^https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs[\d]+/ghostscript-[\d.]+.tar.xz$",
+                )
+                self.assertIsInstance(data.new_version.size, int)
+                self.assertGreater(data.new_version.size, 0)
+                self.assertIsNotNone(data.new_version.checksum)
+                self.assertIsInstance(data.new_version.checksum, str)
+                self.assertNotEqual(
+                    data.new_version.checksum,
+                    "90ed475f37584f646e9ef829932b2525d5c6fc2e0147e8d611bc50aa0e718598",
+                )
+            elif data.filename == "boost_1_74_0.tar.bz2":
+                self.assertIsNotNone(data.new_version)
+                self.assertIsInstance(data.new_version, ExternalFile)
+                self.assertRegex(
+                    data.new_version.url,
+                    r"^https://dl\.bintray\.com/boostorg/release/[\d.]+/source/boost_[\d]+_[\d]+_[\d]+.tar.bz2$",
+                )
+                self.assertIsInstance(data.new_version.size, int)
+                self.assertGreater(data.new_version.size, 0)
+                self.assertIsNotNone(data.new_version.checksum)
+                self.assertIsInstance(data.new_version.checksum, str)
+                self.assertNotEqual(
+                    data.new_version.checksum,
+                    "83bfc1507731a0906e387fc28b7ef5417d591429e51e788417fe9ff025e116b1",
+                )
+            elif data.filename == "flatpak-1.8.2.tar.xz":
                 self.assertIsNotNone(data.new_version)
                 self.assertIsInstance(data.new_version, ExternalFile)
                 self.assertRegex(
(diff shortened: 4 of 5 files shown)


src/checkers/anityachecker.py
```diff
@@ -40,15 +40,15 @@ def check(self, external_data):
 
     def _check_data(self, external_data, latest_version):
         url_template = external_data.checker_data["url-template"]
-        latest_url = Template(url_template).substitute(version=latest_version)
+        latest_url = super(AnityaChecker, self)._substitute_placeholders(url_template, latest_version)
```
gasinvein: Just use `self`  here, no need to invoke `super()`.

src/checkers/anityachecker.py
```diff
@@ -40,15 +40,15 @@ def check(self, external_data):
...
 
         self._update_version(
             external_data, latest_version, latest_url, follow_redirects=False
         )
 
     def _check_git(self, external_data, latest_version):
         tag_template = external_data.checker_data["tag-template"]
-        latest_tag = Template(tag_template).substitute(version=latest_version)
+        latest_tag = super(AnityaChecker, self)._substitute_placeholders(tag_template, latest_version)
```
gasinvein: Same here, just use `self`.

src/checkers/anityachecker.py
```diff
@@ -1,7 +1,7 @@
 import logging
 import urllib.request
 import urllib.parse
-from string import Template
+from distutils.version import LooseVersion
```
gasinvein: Probably unused import?

src/checkers/anityachecker.py
```diff
@@ -40,15 +40,19 @@ def check(self, external_data):
 
     def _check_data(self, external_data, latest_version):
         url_template = external_data.checker_data["url-template"]
-        latest_url = Template(url_template).substitute(version=latest_version)
+        latest_url = super(AnityaChecker, self)._substitute_placeholders(
```
gasinvein: No need to use `super()` here, just use `self._substitute_placeholders()`.

src/checkers/anityachecker.py
```diff
@@ -40,15 +40,19 @@ def check(self, external_data):
...
+        )
 
         self._update_version(
             external_data, latest_version, latest_url, follow_redirects=False
         )
 
     def _check_git(self, external_data, latest_version):
         tag_template = external_data.checker_data["tag-template"]
-        latest_tag = Template(tag_template).substitute(version=latest_version)
+        latest_tag = super(AnityaChecker, self)._substitute_placeholders(
```
gasinvein: Same here, no need for `super()`.

src/checkers/htmlchecker.py
```diff
@@ -109,6 +108,16 @@ def check(self, external_data):
 
         self._update_version(external_data, latest_version, abs_url)
 
+    def _substitute_placeholders(self, template, version):
```
gasinvein: Remove `self` arg here and add `@staticmethod` decorator, since this method doesn't access `self`.

src/checkers/htmlchecker.py
```diff
@@ -109,6 +108,19 @@ def check(self, external_data):
...
+        number_of_fields = len(version_list)
+        for i in range(number_of_fields):
+            template = template.replace("$version" + str(i), str(version_list[i]))
+        template = template.replace("$version", str(version))
+        template = template.replace("$major", str(version_list[0]))
+        if number_of_fields > 1:
+            template = template.replace("$minor", str(version_list[1]))
+        if number_of_fields > 2:
+            template = template.replace("$patch", str(version_list[2]))
+        return template
```
gasinvein: Better construct some `replacements` dict and use `Template.substitute(**replacements)` once instead of using `str.replace` consequently. The later doesn't take syntax into account and can replace unwanted substrings.

gasinvein (changes requested):

src/checkers/htmlchecker.py
Alexander-Wilms: Should the dict contain the keys version, major, minor and patch and Template.substitute be called after the loop?

src/checkers/anityachecker.py
gasinvein: LooseVersion is not accessed directly from this module, you don't need to import it here.

src/checkers/anityachecker.py
```diff
@@ -40,15 +40,15 @@ def check(self, external_data):
 
     def _check_data(self, external_data, latest_version):
         url_template = external_data.checker_data["url-template"]
-        latest_url = Template(url_template).substitute(version=latest_version)
+        latest_url = super()._substitute_placeholders(url_template, latest_version)
```
gasinvein:
```suggestion
        latest_url = self._substitute_placeholders(url_template, latest_version)
```

src/checkers/anityachecker.py
```diff
@@ -40,15 +40,15 @@ def check(self, external_data):
...
 
         self._update_version(
             external_data, latest_version, latest_url, follow_redirects=False
         )
 
     def _check_git(self, external_data, latest_version):
         tag_template = external_data.checker_data["tag-template"]
-        latest_tag = Template(tag_template).substitute(version=latest_version)
+        latest_tag = super()._substitute_placeholders(tag_template, latest_version)
```
gasinvein:
```suggestion
        latest_tag = self._substitute_placeholders(tag_template, latest_version)
```

src/checkers/htmlchecker.py
gasinvein: Yes, something like that, after loop `replacements` should become e.g. `{"major": version_list[0], "minor": version_list[1]}` and so on.

src/checkers/htmlchecker.py
```diff
@@ -109,6 +108,20 @@ def check(self, external_data):
 
         self._update_version(external_data, latest_version, abs_url)
 
+    @staticmethod
+    def _substitute_placeholders(template, version):
+        version_list = LooseVersion(version).version
+        number_of_fields = len(version_list)
+        for i in range(number_of_fields):
```
gasinvein, line 115: You can use `enumerate` here to get both the value and its index, e.g.
```suggestion
        for i, ver in enumerate(version_list):
```

src/checkers/htmlchecker.py
```diff
@@ -109,6 +108,20 @@ def check(self, external_data):
...
+            template = template.replace("$version" + str(i), str(version_list[i]))
```
gasinvein: You can use `enumerate` here to get both the value and its index, e.g.
```suggestion
        for i, ver in enumerate(version_list):
            template = template.replace("$version" + str(i), str(ver))
```
Also calling `len()` is generally cheap for static lists, not need to put in a local variable.

src/checkers/htmlchecker.py
gasinvein: I think something like this should do (you'll need to add string.Template import back):
```suggestion
        tmpl = Template(template)
        tmpl_vars = {"version": version}
        for i, version_part in enumerate(version_list):
            tmpl_vars[f"version{i}"] = version_part
            if i == 0:
                tmpl_vars["major"] = version_part
            elif i == 1:
                tmpl_vars["minor"] = version_part
            elif i == 2:
                tmpl_vars["patch"] = version_part
        return tmpl.substitute(**tmpl_vars)
```

src/checkers/htmlchecker.py
gasinvein: See snippet above on using `enumerate()`

src/checkers/htmlchecker.py
```diff
@@ -109,6 +109,24 @@ def check(self, external_data):
...
+
+        replacements = {"version": version, "major": version_list[0]}
+
+        for i in range(number_of_fields):
+            replacements["version" + str(i)] = str(version_list[i])
+
+        if number_of_fields > 1:
+            template = template_obj.safe_substitute({"minor": str(version_list[1])})
+            if number_of_fields > 2:
+                template = template_obj.safe_substitute({"patch": str(version_list[1])})
```
gasinvein: `template` variable goes nowhere, substituting the template here is pointless. Assign `replacements` entries instead.

src/checkers/htmlchecker.py
```diff
@@ -109,6 +109,24 @@ def check(self, external_data):
...
+    @staticmethod
+    def _substitute_placeholders(template_string, version):
+        version_list = LooseVersion(version).version
+        number_of_fields = len(version_list)
+        template_obj = Template(template_string)
+
+        replacements = {"version": version, "major": version_list[0]}
+
+        for i in range(number_of_fields):
+            replacements["version" + str(i)] = str(version_list[i])
```
gasinvein: Use f-strings instead of concatenation, i.e. `f"version{i}"`

Alexander-Wilms: Yeah, I just saw your suggestion.

gasinvein (commented):
LGTM, thanks for implementing this. Can you please add a test case for it? I think adding yet another module to `org.x.xeyes.yml` that would use the complex template substitution would be enough.

gasinvein (approved):

Alexander-Wilms:
Added test cases. Thanks a lot for the help.

gasinvein:
I think anytia-based test modules should go to anytia test. Also you need to actually run checks on these modules, i.e. extend `tests/test_anityachecker.py` and compare urls we've got for the sample sources to some regex of what is expected.

tests/org.x.xeyes.yml
```diff
@@ -65,3 +65,23 @@ modules:
...
+
+  - name: boost
+    sources:
+      - type: archive
+        url: https://dl.bintray.com/boostorg/release/1.75.0/source/boost_1_75_0.tar.bz2
+        sha256: "0000000000000000000000000000000000000000000000000000000000000000"
+        x-checker-data:
+          type: anitya
+          project-id: 6845
+          url-template: https://dl.bintray.com/boostorg/release/$version/source/boost_$major_$minor_$patch.tar.bz2
```
gasinvein: Template includes `_` into placeholder name and tries to substitute `$major_`. You need to put it in braces, like `${major}_`

Alexander-Wilms:
Should I just add boost and ghostscript modules to org.flatpak.Flatpak.yml and add conditions for "boost_1_75_0.tar.bz2" and "ghostscript-9.26.tar.xz"?

And keep test cases using the new placeholders and the HTML checker in org.x.xeyes.yml?

gasinvein:
> Should I just add boost and ghostscript modules to org.flatpak.Flatpak.yml and add conditions for "boost_1_75_0.tar.bz2" and "ghostscript-9.26.tar.xz"?

Yeah, this should do. Don't mind the manifest filename.

gasinvein:
Sorry, I guess I've misunderstood you previous comment. The newly-added test modules should go into `org.flatpak.Flatpak.yml`, not `org.x.xeyes.yml` (meeh we should rename them all to avoid confusion).

tests/test_anytiachecker.py
```diff
@@ -18,6 +18,36 @@ def test_check(self):
 
         self.assertEqual(len(ext_data), 2)
```
gasinvein: This needs to be incremented by 2 (number of modules you've added).

tests/org.flatpak.Flatpak.yml
```diff
@@ -21,3 +21,24 @@ modules:
           type: anitya
           project-id: 6377
           url-template: https://github.com/flatpak/flatpak/releases/download/$version/flatpak-$version.tar.xz
+          
+  - name: boost
+    sources:
+      - type: archive
+        url: https://dl.bintray.com/boostorg/release/1.74.0/source/boost_1_74_0.tar.bz2
+        sha256: "0000000000000000000000000000000000000000000000000000000000000000"
```
gasinvein: Either set this to actual checksum here or to all zeros in the test assertion, otherwise the assertion always succeeds. Same goes for ghostscript.

tests/test_anytiachecker.py
```diff
@@ -16,8 +16,38 @@ def test_check(self):
...
+                )
+                self.assertIsInstance(data.new_version.size, int)
+                self.assertGreater(data.new_version.size, 0)
+                self.assertIsNotNone(data.new_version.checksum)
+                self.assertIsInstance(data.new_version.checksum, str)
+                self.assertNotEqual(
+                    data.new_version.checksum,
+                    "90ed475f37584f646e9ef829932b2525d5c6fc2e0147e8d611bc50aa0e718598",
+                )
+            if data.filename == "boost_1_74_0.tar.bz2":
```
gasinvein: Should be `elif` instead of `if`

tests/test_anytiachecker.py
```diff
@@ -16,8 +16,38 @@ def test_check(self):
...
+                )
+                self.assertIsInstance(data.new_version.size, int)
+                self.assertGreater(data.new_version.size, 0)
+                self.assertIsNotNone(data.new_version.checksum)
+                self.assertIsInstance(data.new_version.checksum, str)
+                self.assertNotEqual(
+                    data.new_version.checksum,
+                    "83bfc1507731a0906e387fc28b7ef5417d591429e51e788417fe9ff025e116b1",
+                )
             if data.filename == "flatpak-1.8.2.tar.xz":
```
gasinvein: Same here, `elif`.

gasinvein:
LGTM, thanks.

PR #181, opened 2021-05-19
gitchecker: Support semver version scheme for sorting

### LINKED ISSUE #180: Supporting for semantic versioning (SemVer)
In Pika Backup I use semantic versioning. I use Flathub beta channel and want the newest version (beta or release) to be deployed there. However "precedence" in semantic versioning has a twist: `1.0.0-alpha < 1.0.0` while usual sort algorithms come up with the exact opposite result. This means that currently flatpak-external-data-checker prefers 1.0.0-beta over the release version 1.0.0.

Maybe it would be possible to add a flag to enable semantic versioning?

gasinvein:
Fixes #180

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,7 @@ RUN apt-get update && \
       python3-setuptools \
       python3-toml \
       python3-pyelftools \
+      python3-semver \
       squashfs-tools \
       ssh-client \
       jq \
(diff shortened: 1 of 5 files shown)


sophie-h:
Also seems to work out in the wild!

gasinvein:
Ok, this shouldn't affect existing apps anyway.

Issue #169, opened 2021-04-21
Stop checking file/archive sources without x-checker-data?


gasinvein:
Currently we download all the sources specified in the manifest and compare checksums. But is this really needed?
`extra-data` type sources need to be periodically validated, 'cause the checksum change will result in non installable apps. But this doesn't apply to "normal" sources that are downloaded at build time.
From what I can tell, validating "normal" sources mostly causes problems, propagating check error like #168 and #142 to apps that didn't even ask to be checked.
So, I think we should consider limiting implicitly enabled validation to extra-data type sources only.
This would also significantly reduce amount of data to download.
@wjt @barthalion what do you think?

wjt:
That's probably fine – it does seem like it causes more trouble than it saves.

barthalion:
Yeah, I don't think it's really needed.

PR #182, opened 2021-05-19
urlchecker: Add option to strip query parameters

### LINKED ISSUE #179: rotating-urls checker adds information to url which is unneeded
As seen on https://github.com/flathub/io.atom.Atom/pull/101 , it seems that f-e-d-c adds information to the end of the URL.

gasinvein:
If user has requested to strip query parameters, instead of just dropping them, we first traverse HTTP redirects via HEAD requests, then drop query from the latest URL and try to download the resulting URL. This makes sure we calculate checksum for the same file as flatpak-builder will download.
Fixes #179


src/checkers/urlchecker.py
```diff
@@ -76,13 +76,21 @@ async def check(self, external_data):
         else:
             url = external_data.current_version.url
 
+        strip_query = external_data.checker_data.get("strip-query", False)
```
wjt, line 96: I have a theory that, if we scanned all Flathub repos for URLs, we would find 0 examples of query parameters that matter. :)

But I guess false is the conservative default. Should be documented in README though.

gasinvein: Added to readme, please check if the wording is ok.
>if we scanned all Flathub repos for URLs, we would find 0 examples of query parameters that matter

Well, that's actually doable. If your theory proves right, we can just invert the default value.

wjt: I happened to have clones of all Flathub repos from about 6 weeks ago because I used them to prepare that talk.

I used the following program to scrape all the URLs:

```python3
#!/usr/bin/env python3
#
# dump-urls: Extract external URLs from the given manifests
#
# Copyright © 2021 Endless Mobile, Inc.
#
# This program is free software; you can redistribute it and/or modify
# it under the terms of the GNU General Public License as published by
# the Free Software Foundation; either version 2 of the License, or
# (at your option) any later version.
#
# This program is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
# GNU General Public License for more details.
#
# You should have received a copy of the GNU General Public License along
# with this program; if not, write to the Free Software Foundation, Inc.,
# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.

import argparse
import sys
from pathlib import Path

exe = Path(sys.argv[0]).resolve()
sys.path.insert(0, str(exe.parent))

from src.checker import ManifestChecker

def find_manifests(manifest_dirs):
    for manifest_dir in manifest_dirs:
        for extension in (".json", ".yaml", ".yml"):
            manifest_path = (manifest_dir / (manifest_dir.name + extension))
            if manifest_path.is_file():
                yield manifest_path
                break

if __name__ == "__main__":
    parser = argparse.ArgumentParser(
        description="Extract remote URLs from a list of Flathub repos",
    )
    parser.add_argument(
        "manifest_dirs", nargs="*", type=Path,
        help="path to a Flathub repo clone",
    )
    args = parser.parse_args()

    for manifest_path in find_manifests(args.manifest_dirs):
        checker = ManifestChecker(str(manifest_path))
        for x in checker.get_external_data():
            if x.type != x.Type.GIT:
                url = x.current_version.url
                if '?' in url:
                    print(manifest_path.name, url)
```

Here is the output:

```
com.github.tenderowl.norka.json https://downloads.sourceforge.net/project/gtkspell/3.0.10/gtkspell3-3.0.10.tar.xz?r=https%3A%2F%2Fsourceforge.net%2Fprojects%2Fgtkspell%2Ffiles%2F3.0.10%2Fgtkspell3-3.0.10.tar.xz%2Fdownload&ts=1572946840
```

And I confirmed that the URL redirects to the same place with or without the query string. So indeed I think the default can be the other way around.

gasinvein: Noted, thanks for checking this. Still I'm hesitant to change the current behavior, especially for an implicit checker. Maybe flip the default after resolving #169, so that less apps would be affected by potential misbehavior?

README.md
```diff
@@ -128,6 +128,11 @@ The version number for the latest version can be detected in two ways:
   version. (This follows the convention used by
   [`debian/watch`](https://wiki.debian.org/debian/watch) files.)
 
+Some upstream vendors may add unwanted GET query parameters to
+the download URL, e.g. indentifiers for counting unique downloads.
+This may result in URL change without change in the actual data.
+To remove GET query, set `strip-query` proprety to `true`.
```
wjt:
```suggestion
To remove GET query parameters, set `strip-query` property to `true`.
```

README.md
```diff
@@ -128,6 +128,11 @@ The version number for the latest version can be detected in two ways:
   version. (This follows the convention used by
   [`debian/watch`](https://wiki.debian.org/debian/watch) files.)
 
+Some upstream vendors may add unwanted GET query parameters to
+the download URL, e.g. indentifiers for counting unique downloads.
```
wjt:
```suggestion
the download URL, such as identifiers for counting unique downloads.
```

wjt (approved):
OK

src/checkers/urlchecker.py
wjt: #169 is resolved so shall we change the default to "yes, strip the query part" and land this? I really value the default behaviour being what most users want to minimise the amount of gunk they have to blindly paste into their manifest, and to date we have seen many cases where this behaviour is desirable and none where it is not.

gasinvein: If make stripping query the default, ~~maybe we should do it for all checkers, not just the rotating-url one?~~
No, I guess we shouldn't. If we always traverse redirects first, then strip the query - even GitHub downloads will break.

gasinvein:
It appears that we can't test incomplete read if the query is stripped by default, because of
`https://httpbin.org/response-headers?Content-Length=1000`

wjt:
Can't that test set the flag to `false`?

gasinvein:
Not really. These sources are in "implicit check mode", i.g. there is no `x-checker-data`. To set `strip-query: false`, we'll need to redo them in "explicit check mode".

src/checkers/urlchecker.py
```diff
@@ -92,9 +93,17 @@ async def check(self, external_data: ExternalBase):
...
+
         version_string = None
 
         try:
+            if strip_query:
+                async with self.session.head(
+                    url, allow_redirects=True, headers=HTTP_CLIENT_HEADERS
+                ) as head:
+                    url = str(head.url.with_query(""))
```
wjt, line 105: Why is it necessary to traverse redirects?

gasinvein: Because the initial URL without a query can redirect us to an URL with a query, and we want to make sure it works with the query removed.

wjt: I don't quite follow, but I'll take your word for it :)

gasinvein: Well, for example, `http://example.com/download/file.tar.gz` can redirect to `http://cdn.example.com/pub/file.tar.gz?id=key`. Since we're stripping the query from the final url, we should check that `http://cdn.example.com/pub/file.tar.gz` works without `id=key`. Otherwise, we can end up submitting an update with broken URL (in case it doesn't work without query).

PR #183, opened 2021-05-21
Add checker for RPM repositories

### LINKED ISSUE #170: Support checking RPM repositories
We already can pull updates from DEB repos, and it would be nice to support RPM repos as well.
Some apps do host both repo kinds, but since flatpak-builder naively supports extracting only .rpm packages, in some cases pulling updates from an RPM repo may be more convenient.

gasinvein:
Fixes #170

### DIFF
--- a/src/checkers/__init__.py
+++ b/src/checkers/__init__.py
@@ -1,4 +1,5 @@
 from .debianrepochecker import DebianRepoChecker
+from .rpmrepochecker import RPMRepoChecker
 from .jsonchecker import JSONChecker
 from .urlchecker import URLChecker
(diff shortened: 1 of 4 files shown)


wjt (commented):
This looks plausible but, is there no library that can be used from python to avoid having to open code the XML stuff?

gasinvein:
I've searched the internet a bit, but didn't find anything well-established and present in Debian repos. Finding some rpm-md client that supports async IO is even less likely.

wjt (approved):

wjt:
Tests would be nice but I suppose not essential

wjt (approved):

PR #184, opened 2021-05-31
chromiumchecker: Add checker for Chromium components

refi64:
This is needed to be able to auto-update the versions for the LLVM toolchain components in sync with Chromium itself. It also can update the Chromium tarball; the Anitya checker could already technically do this, but this code needs to parse the versions anyway, so actually updating the tarball in sync is a minor addition.

### DIFF
--- a/src/checkers/chromiumchecker.py
+++ b/src/checkers/chromiumchecker.py
@@ -1,0 +1,186 @@
+import aiohttp
+import base64
+import logging
+import re
+import typing as t
+
+from ..lib.externaldata import (
+    Checker,
+    ExternalBase,
+    ExternalData,
+    ExternalDataSource,
+    ExternalGitRepo,
+    ExternalGitRef,
+)
+from ..lib.utils import get_extra_data_info_from_url
+
+log = logging.getLogger(__name__)
+
+
+class Component:
+    NAME: str
+    DATA_CLASS: t.Type[ExternalBase]
+
+    def __init__(
+        self,
+        session: aiohttp.ClientSession,
+        external_data: ExternalBase,
+        latest_version: str,
+    ) -> None:
+        self.session = session
+        self.external_data = external_data
+        self.latest_version = latest_version
+
+        assert latest_version is not None
+
+    async def check(self) -> None:
+        raise NotImplemented
+
+    async def update_external_source_version(self, latest_url):
+        assert latest_url is not None
+
+        try:
+            new_version = await get_extra_data_info_from_url(latest_url, self.session)
+        except (
+            aiohttp.ClientError,
+            aiohttp.ServerConnectionError,
+            aiohttp.ServerDisconnectedError,
+            aiohttp.ServerTimeoutError,
+        ) as e:
+            log.warning("%s returned %s", latest_url, e)
+            self.external_data.state = ExternalData.State.BROKEN
+        else:
+            new_version = new_version._replace(  # pylint: disable=no-member
+                version=self.latest_version
+            )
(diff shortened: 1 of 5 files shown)


refi64:
Whoops did not realize mypy is run in CI now, will fix next chance I get.

gasinvein:
The `Optional[ClientSession]` errors look like my mistake unnoticed until now. I guess `Checker.session` attribute shouldn't be optional, since we always set it up in `__aenter__`.

src/checkers/chromiumchecker.py
```diff
@@ -0,0 +1,183 @@
...
+            result = await response.text()
+
+        update_py = base64.b64decode(result).decode("utf-8")
+        revision = self._CLANG_REVISION_RE.search(update_py).group(1)
+        sub_revision = self._CLANG_SUB_REVISION_RE.search(update_py).group(1)
+
+        return LLVMVersionUtilsMixin.Version(revision, sub_revision)
+
+
+class LLVMGitComponent(Component, LLVMVersionUtilsMixin):
```
gasinvein: Maybe use one more level of inheritance instead of the mixin, e.g. `Component -> LLVMComponent -> LLVMGitComponent`? This will also make Mypy happier.

src/checkers/chromiumchecker.py
```diff
@@ -0,0 +1,183 @@
...
+    async def check(self, external_data: t.Union[ExternalData, ExternalGitRepo]):
+        assert self.should_check(external_data)
+
+        component_name = external_data.checker_data.get(
+            "component", ChromiumComponent.NAME
+        )
+        if component_name not in self._COMPONENTS:
+            log.error(f"Invalid component: {component_name}")
+            external_data.state = ExternalBase.State.BROKEN
+            return
```
gasinvein: Most other checkers consider malformed metadata fatal, so if the user provided invalid `component` - it's ok to raise exception. So, I think this check is redundant.

refi64:
@gasinvein should I just ignore the `session` access errors for now then?

gasinvein:
Pushed fix to master, please rebase.

refi64:
@gasinvein pushed, static type checks are green now!

src/checkers/chromiumchecker.py
```diff
@@ -0,0 +1,186 @@
...
+        return result[0]["versions"][0]["current_version"]
+
+    async def check(self, external_data: t.Union[ExternalData, ExternalGitRepo]):
+        assert self.should_check(external_data)
+
+        component_name = external_data.checker_data.get(
+            "component", ChromiumComponent.NAME
+        )
+        if component_name not in self._COMPONENTS:
+            raise ValueError(f"Invalid component {component_name}")
```
gasinvein, line 178: I guess this is redundant, too: we'll get a `KeyError` from the next statement anyway if invalid component specified.

refi64: It sort of is, but I figured this would be a bit clearer / require less knowledge of Python to understand.

gasinvein: Well, I assume there won't be other users of this checker besides Chromium flatpak maintainers.

gasinvein:
This looks good overall if it works for you. Don't mind writing a test case?

refi64:
@gasinvein tests should be added now, I did lower the logging level because the Chromium files are large enough that running the checker absolutely drowns the console in tons of asyncio debugging information

gasinvein (approved):
LGTM, thanks.

Issue #168, opened 2021-04-21
Sporadic update PRs with invalid checksums for SourceForge URLs


Zlopez:
Hi,

today the flathubbot opened a strange [PR](https://github.com/flathub/org.gnome.Chess/pull/28) on org.gnome.Chess. PR is opened for dependency which doesn't have `x-data-checker` options set up and it just changed the hash without changing the URL. I looked at the URL and the file had last update in 2016.

I will close the PR for now, because the build is failing anyway.

gasinvein:
It looks much like #142, but I don't know why it pops up again. Unlike `requests`, `aiohttp` does check content-length by default.

gasinvein:
There are multiple other apps hit by this at the same time, and seems like all affected sources are hosted on SourceForge.
https://github.com/flathub/org.octave.Octave/pull/142
https://github.com/flathub/org.gnome.Evince/pull/47
https://github.com/flathub/org.gnome.Chess/pull/28
https://github.com/flathub/org.gnome.Books/pull/26
https://github.com/flathub/org.freedesktop.Sdk.Extension.texlive/pull/45
https://github.com/flathub/org.flightgear.FlightGear/pull/39
https://github.com/flathub/net.gwyddion.Gwyddion/pull/23
https://github.com/flathub/me.kozec.syncthingtk/pull/34
https://github.com/flathub/io.github.peazip.PeaZip/pull/5
https://github.com/flathub/im.riot.Riot/pull/175
https://github.com/flathub/com.sweethome3d.Sweethome3d/pull/22
https://github.com/flathub/com.chez.GrafX2/pull/5
And the invalid new checksum is the same one everywhere: `251a85b3bac687974f360d3796048c20ded3bf0bd69e0d1cfd1db23d013f89ed`

Vanadiae:
Thanks @gasinvein for mentioning the affected MRs here :)

gasinvein:
Any idea what has happened? Probably some SourceForge issue? Looking at flathub [download cache](https://sources.flathub.org/sources/downloads/251a85b3bac687974f360d3796048c20ded3bf0bd69e0d1cfd1db23d013f89ed/), it seems like this has occurred before.

SISheogorath:
Source forge seems to have no permalinks to source code and/or employing some sort of blocking which results in the external data checker getting confused.

gasinvein:
[...]

PR #186, opened 2021-06-05
Disable implicit check for non-extra-data

### LINKED ISSUE #169: Stop checking file/archive sources without x-checker-data?

gasinvein:
Closes #169, should reduce the impact of #168.


gasinvein:
Well, I guess this better be merged until sourceforge strikes again.

PR #187, opened 2021-06-06
chromiumchecker: Use NotImplementedError exception

lionirdeadman:
NotImplemented is not an exception
This was caught by https://lgtm.com/projects/g/flathub/flatpak-external-data-checker/alerts/


wjt:
Nice catch. Is there a Lgtm GitHub action we could add to catch future issues before merge?

PR #200, opened 2021-07-09
Guess if external data is broken

### LINKED ISSUE #193: Implicit automerge doesn't trigger for most checkers
It used to work before. I suspect that it regressed after #119. More specifically, 1658c4190746b0adf44b8fc4daebc11caa827fef makes it so that if if an `extra-data` source has check method other than url, the data never reaches `URLChecker` and never gets `BROKEN` status, thus, the condition for implicit automerge is never met.
CC @Garbee

gasinvein:
This adds logic to check if a new version is an actual new upstream release, or it's the same version has changed.
For file sources, only compare urls: if they match, assume it's the same release.
For git sources, compare tag and branch in addition to urls.
If the new_version retrieved by checker is the same release, but integrity (checksum or commit) doesn't match, mark the data as broken.
Should fix #193

### DIFF
--- a/src/lib/externaldata.py
+++ b/src/lib/externaldata.py
@@ -102,9 +102,12 @@ class ExternalBase(abc.ABC):
         raise NotImplementedError
 
     def set_new_version(
-        self, new_version: t.Union[ExternalFile, ExternalGitRef], is_update=True
+        self, new_version: t.Union[ExternalFile, ExternalGitRef], is_update=None
     ):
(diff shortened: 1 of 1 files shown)


gasinvein:
@barthalion @wjt I'm concerned about high-level logic here. Is it correct assumption that a new version is a new upstream release when it's URL has changed? Maybe we should take something else into account?

barthalion:
Looks good to me.

PR #207, opened 2021-07-25
gitchecker: Don't crash when no tags match

refi64:


refi64:
Looks like the final test was just some transient network issues, but I don't seem to have permission to restart it.

gasinvein:
Meeh, I was writing a comment asking to do this in try/except manner, but that's not a big deal.

PR #208, opened 2021-07-25
Support different checksums

gasinvein:
This is rather big change-set for a seemingly small feature. Given that different checkers can get different digests, while sources may set different digests and expect them to be updated, we need some layer of abstraction between checkers and flatpak-builder sources. Thus, internal representation of checksum changed from plain sha256 digests to a special `MultiDigest` container.

While retrieving remote checksums is straightforward, applying updates to sources isn't. There are some corner cases, e.g.

- What should we do if the checker got `md5` and `sha256`, but the source has set only `sha512`?

### DIFF
--- a/src/checkers/debianrepochecker.py
+++ b/src/checkers/debianrepochecker.py
@@ -42,6 +42,7 @@ import apt
 import apt_pkg
 
 from ..lib.externaldata import Checker, ExternalFile, ExternalBase
+from ..lib.checksums import MultiDigest
 from ..lib.utils import get_timestamp_from_url
 
 apt_pkg.init()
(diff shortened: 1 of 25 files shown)


src/checkers/debianrepochecker.py
```diff
@@ -57,10 +63,26 @@
...
+    "MD5Sum": "md5",
+}
 
 LOG = logging.getLogger(__name__)
 
 
+def read_deb_hashes(deb_hashes: apt_pkg.HashStringList) -> MultiDigest:
+    digests: t.Dict[str, str] = {}
+    deb_hash: apt_pkg.HashString
+    for deb_hash in deb_hashes:
```
gasinvein: Any idea why Mypy complains? `apt_pkg` is an extension module, but Mypy doesn't complain on my machine.
UPD: Silenced with `# type: ignore`, but the question remains.

gasinvein:
I've realized that checksum handling is not really tied to flatpak-builder sources and moved the new classes to a separate module.

gasinvein:
I think this should be mostly fine as it is. At least I didn't come up with major changes since this PR was opened.

src/lib/checksums.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class MultiDigest(t.NamedTuple):
+    md5: t.Optional[str] = None
+    sha1: t.Optional[str] = None
+    sha256: t.Optional[str] = None
+    sha512: t.Optional[str] = None
+
+    @classmethod
+    def form_source(cls, source: t.Dict):
```
wjt: Sorry to nitpick but should this be `from_source`? and `-> MultiDigest`?

wjt (approved):
This is surprisingly non-invasive. I'm really sorry to leave a review whose only suggestion is a typo in the first commit…

src/lib/checksums.py
gasinvein: Right. And the code autocomplete made sure that the typo got successfully propagated everywhere.

gasinvein:
> This is surprisingly non-invasive

It's rather invasive on the DebianRepoChecker. I hope we'll find a better way to get all the checksums from APT in the future.

wjt:
True.

PR #218, opened 2021-08-06
Collect errors and set non-zero exit code if any occured

### LINKED ISSUE #201: Provide an option to only validate x-checker-data
- [x] Some option like `--validate-only` that would cause f-e-d-c to only check if `x-checker-data` is valid, without doing actual updates and exiting with non-zero status only when an error occurred, could be useful to integrate f-e-d-c as part of Flathub CI.
- [ ] Preferably, f-e-d-c in `--validate-only` mode shouldn't hit the network, but it would require a bit more work (we'll need to split some `load_checker_data()` method from `check()` in each checker). But this can be done afterwards.

gasinvein:
Collect all the non-fatal errors occurred during check, and if a special option is given, set the exit status based on amount of errors collected.
Fixes #201

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -30,6 +30,8 @@ import os
 import subprocess
 import sys
 import asyncio
+from enum import IntFlag
+import typing as t
 
 from github import Github
 
@@ -102,6 +104,12 @@ def print_outdated_external_data(manifest_checker: checker.ManifestChecker):
         )
         print(message, flush=True)
     return len(ext_data)
+
+
+def print_errors(manifest_checker: checker.ManifestChecker) -> int:
+    # TODO: Actually do pretty-print collected errors
+    errors = manifest_checker.get_errors()
+    return len(errors)
 
 
 def check_call(args):
@@ -285,8 +293,8 @@ def parse_cli_args(cli_args=None):
         action="store_true",
     )
     parser.add_argument(
-        "--validate-only",
-        help="Do not apply updates, only check if metadata is valid",
+        "--check-outdated",
+        help="Exit with non-zero status if outdated sources were found and not updated",
         action="store_true",
     )
     parser.add_argument(
@@ -298,8 +306,11 @@ def parse_cli_args(cli_args=None):
     return parser.parse_args(cli_args)
 
 
-async def run_with_args(args):
+async def run_with_args(args: argparse.Namespace) -> t.Tuple[int, int, bool]:
     init_logging(logging.DEBUG if args.verbose else logging.INFO)
+
+    should_update = args.update or args.commit_only or args.edit_only
+    did_update = False
 
     manifest_checker = checker.ManifestChecker(args.manifest)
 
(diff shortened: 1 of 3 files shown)


src/main.py
```diff
@@ -285,8 +292,8 @@ def parse_cli_args(cli_args=None):
         action="store_true",
     )
     parser.add_argument(
-        "--validate-only",
-        help="Do not apply updates, only check if metadata is valid",
+        "--validate",
+        help="Set exit code based on errors amount",
```
wjt: Why is this conditional? Why not always exit with a defined error code in this case? (Sincere question.)

```
         help="Exit with non-zero status if errors are found while checking",
```

gasinvein: Currently we set exit code based on amount of outdated data - 0 if all up to date and 1 if something is outdated. I suppose we should retain this default behaviour?

gasinvein: Maybe we should combine non-zero exit code? E.g. `1` if outdated, `2` if errors, `3` if both.

gasinvein: Now this is unconditional.

gasinvein:
@wjt Will this work for Endless' CI, given the default behaviour has changed? I guess we may need to add the new `--check-outdated` option to the wrapper?

src/main.py
```diff
@@ -298,41 +306,50 @@ def parse_cli_args(cli_args=None):
...
-                        open_pr(
-                            subject, body, branch, manifest_checker=manifest_checker
-                        )
-                return (outdated_num, True)
+    return outdated_num, errors_num, did_update
 
-            log.warning("Can't automatically fix any of the above issues")
 
-    return (outdated_num, False)
+class ResultCode(IntFlag):
```
wjt, line 341: I have never seen IntFlag before. Nice!

src/main.py
```diff
@@ -298,41 +306,50 @@ def parse_cli_args(cli_args=None):
...
+    SUCCESS = 0
+    ERROR = 1
+    OUTDATED = 2
 
 
 def main():
+    res = ResultCode(0)
```
wjt: Why not:

```suggestion
    res = ResultCode.SUCCESS
```

?

src/main.py
```diff
@@ -104,6 +106,12 @@ def print_outdated_external_data(manifest_checker: checker.ManifestChecker):
     return len(ext_data)
 
 
+def print_errors(manifest_checker: checker.ManifestChecker):
```
```suggestion
def print_errors(manifest_checker: checker.ManifestChecker) -> int:
```

A little weird that a function called `print_*` returns a value and doesn't print :-)

wjt (approved):
Looks good. I don't think the Endless CI needs to change.

src/main.py
gasinvein: It returns an `int` just like `print_outdated_external_data()` does.
I didn't yet figure out how to print errors in a readable and useful manner.

src/main.py
gasinvein: Very bitwise, much C

PR #219, opened 2021-08-11
Check content type of downloaded files

### LINKED ISSUE #168: Sporadic update PRs with invalid checksums for SourceForge URLs

gasinvein:
Add a list of known-wrong `Content-Type` header patterns, and reject downloaded file if its `Content-Type` matches. The idea is to catch error pages returned by services without HTTP error status code, and avoid submitting them as "updates".
Should fix #168, assuming that the infamous SourceForge page has C-T alike `text/html`.

Ideally, we should add C-T denylist for URLChecker, too, but just throwing it in will break our tests (httbin[go] returns `text/html` for its `/base64/` URL). Given that it's most primarily targeted at SourceForge, and sources from there are unlikely to be `extra-data`, I consider the denylist for URLChecker not as important.

### DIFF
--- a/src/checkers/htmlchecker.py
+++ b/src/checkers/htmlchecker.py
@@ -162,8 +162,8 @@ class HTMLChecker(Checker):
             )
         except NETWORK_ERRORS as err:
             raise CheckerFetchError from err
-        else:
-            new_version = new_version._replace(  # pylint: disable=no-member
-                version=latest_version
-            )
-            external_data.set_new_version(new_version)
+
+        new_version = new_version._replace(
+            version=latest_version  # pylint: disable=no-member
+        )
+        external_data.set_new_version(new_version)
--- a/src/checkers/htmlchecker.py
+++ b/src/checkers/htmlchecker.py
@@ -25,7 +25,12 @@ from string import Template
 from distutils.version import LooseVersion
 import typing as t
 
-from ..lib import utils, NETWORK_ERRORS
+from ..lib import (
+    utils,
+    NETWORK_ERRORS,
+    WRONG_CONTENT_TYPES_FILE,
+    WRONG_CONTENT_TYPES_ARCHIVE,
+)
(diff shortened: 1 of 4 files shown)


nanonyme:
LGTM

gasinvein:
Should we add more wrong-C-T patterns? Or make current ones more precise?

wjt:
I think it is okay like this. At the same time, if we are targeting a specific sourceforge error, I think a shasum denylist could also be OK!

nanonyme:
@wjt the idea with MIME was targeting not just sourceforge but any provider who uses HTTP 200 for "down for maintenance" page. It seems this was routinely hit at one point because sourceforge went multiple days into disaster recovery after losing their disk data.

gasinvein:
@wjt Checksum-based denylist would be easier to implement, but as @nanonyme pointed out in matrix chat, an error pages' checksum can change for any reason. And we still have no way of knowing it in advance, so if the checksum does change, we'll have no other option but to wait for #168 to repeat.

nanonyme:
Checksum is fine if you're reluctant to have generic protection. We don't know exactly what the page said. Just that it was something about disaster maintenance.

gasinvein:
So, should we add a checksum denylist, too?

wjt:
The generic approach of checking mime type is probably indeed more future-proof

wjt (approved):
LGTM but now has conflicts :9

wjt (approved):

Justinzobel:
Can we perhaps get a way to whitelist certain URLs or wildcard addresses where we know the file is fine even if the content type doesn't match?

PR #220, opened 2021-08-18
Skip yaml anchors

### LINKED ISSUE #189: Deduplicate external data before checking
I've stumbled upon this issue while adding x-checker-data for the mingw-w64 SDK extension.
Some flatpak manifest have duplicated sources, which is rather common among multilib builds (every source is used twice there). Some flatpaks, e.g. mingw-w64, can have as many as 11 instances of the same source used in different modules.
This can be work-around'ed by placing the repeated source into separate file, but it would be nice if f-e-d-c was smart enough not to check the same external data several times.

gasinvein:
Fixes #189, supersedes and closes #199

### DIFF
--- a/src/checker.py
+++ b/src/checker.py
@@ -67,9 +67,7 @@ def find_appdata_file(directory, appid):
     return None
 
 
-def _external_source_filter(manifest_path: str, source) -> t.Optional[bool]:
-    if not isinstance(source, str):
-        return None
+def _external_source_filter(manifest_path: str, source: str) -> bool:
     source_path = os.path.join(os.path.dirname(manifest_path), source)
     source_size = os.stat(source_path).st_size
     if source_size > MAX_MANIFEST_SIZE:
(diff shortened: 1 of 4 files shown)


src/checker.py
```diff
@@ -173,12 +171,26 @@ def _collect_module_data(self, module_path: str, module: t.Union[str, t.Dict]):
...
+            # Find external sources (path strings) first
+            if isinstance(source, str):
+                if _external_source_filter(module_path, source):
+                    source_paths.append(source)
+                continue
+            assert isinstance(source, dict)
+            # Next, collect only unique sources
+            # NOTE here we rely on ruamel.yaml to make YAML aliases into
+            # pointers to the same dict object ridden from YAML anchor
+            if not any(d.source is source for d in manifest_datas):
```
wjt: O(n²) in the number of sources. Maybe it's fine because 'n' is small.

But I can't quite see how this will work – `manifest_datas` is not updated until after this loop, so if there are aliased sources in the same file, how will this catch them? Oh, because this loop is for sources in a particular module, and `manifest_datas` is for the entire file. So we would not catch duplicates in the case of the same source being aliased in the same module, but that will not happen.

Any idea how to test this? My suggestion would be:

- Manifest using YAML anchors for a source
- Mock checker in the test case which records how many times it has been invoked
- Assert that it is invoked exactly once

gasinvein: >O(n²) in the number of sources. Maybe it's fine because 'n' is small.

I believe `X is Y`, unlike `X == Y`, compares only pointers, not actual objects, so this should be fast enough.

>we would not catch duplicates in the case of the same source being aliased in the same module, but that will not happen

Ah, good point. This is unlikely to happen in practice, yes, but I think this should be fixed anyway. I was thinking of splitting "collect data from source" logic to a separate method, it would be called per-source, so it'll probably will fix this.

>Any idea how to test this?

Just run the checker against a manifest with, say, 2 anchors with 1 alias each (4 sources in total), and verify that we collected exactly 2 data items?

gasinvein: @wjt As the easiest way to test this, I've added an alias to our enrypoint-testing manifest (the test verifies amount of outdated sources), and rearranged this code block in a way that we can see if both of the `any(...)` check outcomes are covered.
Do you think this is sufficient?

gasinvein:
On the performance thing, out of curiosity I did some quick benchmarks, using a generated manifest with 1929 sources (all of them are skipped). Time f-e-d-c took to process the manifest on my machine (average of 100 sequential runs):
- 118 ms with the check removed
- 215 ms with `d.source is source` check
- 291 ms with `d.source == source` check

gasinvein:
Added an anchor/alias pair to a yaml manifest comparison test, in order to verify that we preserve yaml aliases, just in case.

src/checker.py
gasinvein: Now we also directly check amount of data collected in `test_checker.py`

PR #223, opened 2021-08-26
Add --force-[no-]fork CLI parameters

wjt:
I am trying to use this in GitHub Actions, with the default GITHUB_TOKEN
provided to the action. However, origin_repo.permissions.push is False.
Some digging suggests that this may be an artefact of the kind of token
being used here.[0] However, if we just go ahead and push and create the
CR directly to the repo, it works fine.

So, add a --force-no-fork CLI parameter to override the repository
permissions check, and a --force-fork parameter for symmetry. The
default is still to check the repository permissions. (Of course if you
--force-no-fork and you really don't have permission to push or create
PRs, you'll hit an error a few lines later.)

[0] https://github.com/endlessm/eos-google-chrome-app/pull/96

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -171,7 +171,12 @@ AUTOMERGE_DUE_TO_BROKEN_URLS = (
 )
 
 
-def open_pr(subject, body, branch, manifest_checker=None):
+def open_pr(
+    subject: str,
+    body: str,
+    branch: str,
+    manifest_checker: checker.ManifestChecker = None,
+):
     try:
         github_token = os.environ["GITHUB_TOKEN"]
     except KeyError:
--- a/src/main.py
+++ b/src/main.py
@@ -253,7 +253,7 @@ def open_pr(
     for pr in open_prs:
         log.info("Found open PR: %s", pr.html_url)
 
-        if origin_repo.permissions.push and (automerge or force_automerge):
+        if automerge or force_automerge:
             pr_commit = pr.head.repo.get_commit(pr.head.sha)
             if pr_commit.get_combined_status().state == "success" and pr.mergeable:
                 log.info("PR passed CI and is mergeable, merging %s", pr.html_url)
--- a/src/main.py
+++ b/src/main.py
@@ -176,6 +176,7 @@ def open_pr(
     body: str,
     branch: str,
     manifest_checker: checker.ManifestChecker = None,
+    fork: t.Optional[bool] = None,
 ):
     try:
         github_token = os.environ["GITHUB_TOKEN"]
@@ -194,8 +195,14 @@ def open_pr(
     )
     origin_repo = g.get_repo(parse_github_url(origin_url))
 
-    if origin_repo.permissions.push:
-        log.debug("origin repo is writable")
+    if fork is True:
+        log.debug("creating fork (as requested)")
+        repo = user.create_fork(origin_repo)
+    elif fork is False:
+        log.debug("not creating fork (as requested)")
+        repo = origin_repo
+    elif origin_repo.permissions.push:
+        log.debug("origin repo is writable; not creating fork")
         repo = origin_repo
     else:
         log.debug("origin repo not writable; creating fork")
@@ -308,6 +315,33 @@ def parse_cli_args(cli_args=None):
         type=ExternalData.Type,
         choices=list(ExternalData.Type),
     )
+
+    fork = parser.add_argument_group(
+        "control forking behaviour",
+        "By default, %(prog)s pushes directly to the GitHub repo if the GitHub "
+        "token has permission to do so, and creates a fork if not.",
+    ).add_mutually_exclusive_group()
+    fork.add_argument(
+        "--always-fork",
+        action="store_const",
+        const=True,
+        dest="fork",
+        help=(
+            "Always push to a fork, even if the user has write access to the "
+            "upstream repo"
+        ),
+    )
+    fork.add_argument(
+        "--never-fork",
+        action="store_const",
+        const=False,
+        dest="fork",
+        help=(
+            "Never push to a fork, even if this means failing to push to the "
+            "upstream repo"
+        ),
+    )
+
     return parser.parse_args(cli_args)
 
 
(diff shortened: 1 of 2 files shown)


gasinvein:
Hmm, do you think this may be a PyGitHub issue, or just and GitHub API quirk?

wjt:
@dbnicholson did a lot of digging last night and concluded:

> After a bit of fiddling I figured out that the tokens created for Github Actions are Github Apps tokens with the [ghs](https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats/#identifiable-prefixes) prefix. I believe that's incompatible with the way f-e-d-c is checking access since repo permissions are (I believe) tied to actual users. So looking at the `permissions` object in the API `/repos/<repo>` endpoint will always be all `false`.

I checked the PyGitHub code and it doesn't do anything fancy, just turns the dictionary returned into a Python object. So I think it's an API quirk. It does seem unsatisfactory to have to do this though.

gasinvein:
> I believe that's incompatible with the way f-e-d-c is checking access since repo permissions are (I believe) tied to actual users

Oh, so we probably should figure out another way to check our push permission?

src/main.py
```diff
@@ -171,7 +171,7 @@ def commit_changes(changes):
 )
 
 
-def open_pr(subject, body, branch, manifest_checker=None):
+def open_pr(subject, body, branch, manifest_checker=None, force_fork=None):
```
gasinvein: Can you please add type hints to these args?

src/main.py
```diff
@@ -324,7 +345,13 @@ async def run_with_args(args: argparse.Namespace) -> t.Tuple[int, int, bool]:
             with indir(os.path.dirname(args.manifest)):
                 subject, body, branch = commit_changes(changes)
                 if not args.commit_only:
-                    open_pr(subject, body, branch, manifest_checker=manifest_checker)
+                    open_pr(
+                        subject,
+                        body,
+                        branch,
+                        manifest_checker=manifest_checker,
+                        force_fork=args.force_fork,
```
gasinvein: I suspect that `args.force_fork` will default to `False` if neither `--force-fork` nor `--force-no-fork` is given, while we should default to `None`.

dbnicholson:
> > I believe that's incompatible with the way f-e-d-c is checking access since repo permissions are (I believe) tied to actual users
> 
> Oh, so we probably should figure out another way to check our push permission?

I messed around with this for a long time, even going so far as to create a test github app to see what I could do with the token. Unfortunately, I don't think there will be much we can do.

https://github.community/t/retrieve-permissions-as-a-github-app-installation/16125 is someone doing almost the same thing. You can see that when he gets the repo permissions from `/installation/repositories`, they all say `false` even though his token has `"contents":"write"`. OK, so we should just look at the installation token permissions. Unfortunately, they're only returned in the response from when the app creates the token. That's what Github Actions does at the beginning of the workflow and you can see it print them, but AFAICT it doesn't provide them to the job anywhere as an environment variable or similar. There does not appear to be a way to introspect an existing installation token in the Github API.

The best I could come up with is try to figure out if we have an installation token (i.e., Github Actions automated token) or a personal access token and then just assume that it has the permissions to push to the repo and make a PR. Some options:

1. Try fetching `/installation/repositories`. If you have a personal access token, you'll get a 403. Unfortunately, PyGithub gets in the way here as it doesn't provide any means to use that API directly from an installation token. See https://github.com/PyGithub/PyGithub/issues/1776 for some discussion of how it handles Github installations.

2. Just look at the token prefix. If it's `ghs_`, then it's an installation token. If it's anything else (like a newer personal access token prefixed with `ghp_` or an older token that's just 40 hex characters), look at the repo permissions.

gasinvein:
@dbnicholson Thanks for sorting this out and sharing you findings.
Since there is no straightforward way to determine the token permissions, I guess we better explicitly select fork/direct push behaviour, i.e. what this PR adds.

dbnicholson:
One other possibility I found to detect if you have a personal access token or not. After you make an API request, you'll see the OAuth scopes in the [X-OAuth-Scopes](https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps) header. But that only happens for a personal access token. That header isn't returned for an installation token.

In PyGithub, the main `Github` object exposes this via the `oauth_scopes` attribute. It will be a dictionary for a personal access token or `None` for an installation token. But, again, only after you've actually made an API request since it's initially `None`. Something like this (untested):

```
diff --git a/src/main.py b/src/main.py
index 92d798f..78b7e47 100755
--- a/src/main.py
+++ b/src/main.py
@@ -181,7 +181,16 @@ def open_pr(subject, body, branch, manifest_checker=None):
     )
     origin_repo = g.get_repo(parse_github_url(origin_url))
 
-    if origin_repo.permissions.push:
+    if g.oauth_scopes is None:
+        # App installation token (e.g., Github Actions), assume we can push
+        log.debug("app installation token, assuming repo writable")
+        can_push = True
+    else:
+        # Personal access token, use the repo permissions to determine
+        # if we can push
+        can_push = origin_repo.permissions.push
+
+    if can_push:
         log.debug("origin repo is writable")
         repo = origin_repo
     else:
@@ -233,7 +242,7 @@ def open_pr(subject, body, branch, manifest_checker=None):
     for pr in open_prs:
         log.info("Found open PR: %s", pr.html_url)
 
-        if origin_repo.permissions.push and (automerge or force_automerge):
+        if can_push and (automerge or force_automerge):
             pr_commit = pr.head.repo.get_commit(pr.head.sha)
             if pr_commit.get_combined_status().state == "success" and pr.mergeable:
                 log.info("PR passed CI and is mergeable, merging %s", pr.html_url)
```

src/main.py
wjt: I was so confident that this was correct that I wrote a test case and it turns out you are right.

wjt:
Interesting, I like that idea better. You also caught the other access to `origin_repo.permissions` which I missed, and would have made me sad when the next Chrome update didn't get automerged.

The third option is just to make a normal token for a user that has write access to the repo, just like the current Jenkins setup. Slightly more set-up work but would mean no weird heuristics in fedc and also no `--override-random-stuff` flags.

gasinvein:
I think that "do as I say" flags here are useful in any case, since otherwise we rely on some external API that may change.

dbnicholson:
Right, there's also the case of "I want to push to my fork even though I can push to the origin". In some projects I work on even the primary contributors work from personal forks. Without the option you couldn't do that.

gasinvein (approved):
LGTM.

PR #226, opened 2021-08-27
Remove Endless-specific CI scripts

wjt:
We have switched to using the OCI container from GitHub Actions on
the one remaining repository.

https://github.com/endlessm/eos-google-chrome-app/pull/96


wjt:
Draft because I would like to await a successful "real" run of the Action before calling it Done.

PR #230, opened 2021-09-14
Start new http session for downloads without auto decompress

### LINKED ISSUE #159: Encoded content is decoded resulting in wrong size/checksum

gasinvein:
Fixes #159.

### DIFF
--- a/tests/test_checker.py
+++ b/tests/test_checker.py
@@ -23,6 +23,8 @@ import logging
 import os
 import unittest
 import tempfile
+import hashlib
+import gzip
 
 from xml.dom import minidom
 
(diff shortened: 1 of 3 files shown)


tests/test_checker.py
```diff
@@ -437,11 +439,19 @@ async def test_check(self):
             relative_redirect.new_version.url,
             "https://httpbingo.org/base64/MzAtNTAgZmVyYWwgaG9ncyEK",
         )
-        self.assertEqual(
+        # XXX: Gzipped response will have size/checksum different from plain; match any
```
wjt: Isn't the intention of the patch to make these deterministically have the checksum of the compressed content? In which case, why does the test allow both?

gasinvein: We can't tell httpbingo to encode (or not) the content. Currently httbingo feeds us gzipped response, but I won't be surprised if it suddenly starts to feed plain, thus here check for both.

wjt: Can we put this in the comment?

gasinvein: Sure, done.

gasinvein: If you know where we can get a response with well-known body that is deterministically gzipped, we can add it to tests.

wjt: Probably the sustainable fix here is to put an HTTP server into the test suite and serve data locally. e.g. https://pytest-httpserver.readthedocs.io/en/latest/

wjt (approved):

PR #235, opened 2021-09-24
Validate source early

gasinvein:
In case of schema errors in the manifest, f-e-d-c will usually throw an unhandled exception, which may not be easy for the user to figure out what's wrong.
Instead, validate source against a JSON schema as early as possible, and if the schema is invalid - print the error and skip such source.

### DIFF
--- a/src/checker.py
+++ b/src/checker.py
@@ -45,6 +45,8 @@ from .lib.errors import (
     AppdataNotFound,
     AppdataLoadError,
     ManifestLoadError,
+    SourceLoadError,
+    SourceUnsupported,
 )
 
 import logging
@@ -238,8 +240,14 @@ class ManifestChecker:
         manifest_datas = self._external_data.setdefault(source_path, [])
         if any(d.source is source for d in manifest_datas):
             return
-        data = ExternalData.from_source(source_path, source)
-        if data:
+        try:
+            data = ExternalData.from_source(source_path, source)
+        except SourceUnsupported as err:
+            log.debug(err)
+        except SourceLoadError as err:
+            self._errors.append(err)
+            log.error(err)
+        else:
             manifest_datas.append(data)
 
     async def _check_data(
--- a/src/lib/errors.py
+++ b/src/lib/errors.py
@@ -26,6 +26,14 @@ class ManifestUpdateError(ManifestError):
     """Error updating flatpak-builder manifest"""
 
 
+class SourceLoadError(ManifestLoadError):
+    """Error loading flatpak-builder source item"""
+
+
+class SourceUnsupported(SourceLoadError):
+    """Don't know how to handle flatpak-builder source item"""
+
+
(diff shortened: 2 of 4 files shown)


src/lib/externaldata.py
```diff
@@ -67,14 +78,15 @@ class State(Enum):
...
-            data_type = cls.Type(source.get("type"))
-        except ValueError:
+            jsonschema.validate(source, cls.SOURCE_SCHEMA)
+        except jsonschema.ValidationError as err:
+            raise ManifestLoadError("Error reading source") from err
+
+        if not source.get("url"):
             return None
 
+        data_type = cls.Type(source["type"])
```
wjt: Previously, an unrecognised `type` would be caught by `except ValueError`. With this change the `ValueError` will propagate outwards. I don't think we want that because there are many other valid types beyond the four in `class Type(Enum)` above.

gasinvein: I see, this wasn't intentional. Redone this wrapping unsupported source types in custom exception, handled in `MenfestChecker` instead of `ExternalData`.

src/checker.py
```diff
@@ -238,9 +239,16 @@ def _collect_source_data(
...
+        try:
+            data = ExternalData.from_source(source_path, source)
+        except ManifestUnsupported as err:
+            log.debug(err)
+            return
+        except ManifestLoadError as err:
+            self._errors.append(err)
+            log.error(err)
+            return
+        manifest_datas.append(data)
```
wjt: I am personally a fan of this syntax but I know it's an acquired taste.

```suggestion
        try:
            data = ExternalData.from_source(source_path, source)
        except ManifestUnsupported as err:
            log.debug(err)
        except ManifestLoadError as err:
            self._errors.append(err)
            log.error(err)
        else:
            manifest_datas.append(data)
```

src/lib/errors.py
```diff
@@ -22,6 +22,10 @@ class ManifestLoadError(ManifestError):
     """Error loading flatpak-builder manifest"""
 
 
+class ManifestUnsupported(ManifestLoadError):
+    """Don't know how to handle manifest"""
```
wjt: This exception is raised when a *source* is unsupported, not a whole manifest.

wjt (changes requested):

src/lib/errors.py
gasinvein: This was intended as a generic exception when the input data is valid, but fedc doesn't know how to handle it. If limit it to sources, I suppose it should be renamed to e.g. `SourceUnsupported`?

src/checker.py
gasinvein: I usually try to keep next-level-indented blocks as short as possible, or avoid them altogether. Here `except` blocks are 1 or 2 lines, so I guess it makes sense to avoid `return`s.

PR #236, opened 2021-10-01
jsonchecker: Use GitHub token for API queries, if set

gasinvein:
GitHub REST API rate limit is rather low; the json checker can quickly drain it, what is currently happening with shared f-e-d-c instance on Flathub. We can authorize GitHub API queries to significantly increase this limit.

If `GITHUB_TOKEN` env var is set, and GitHub API host detected in JSON data url, add the token to `Authorization` header to authorize API request.

### DIFF
--- a/src/checkers/jsonchecker.py
+++ b/src/checkers/jsonchecker.py
@@ -3,6 +3,9 @@ import re
 from datetime import datetime
 import typing as t
 import subprocess
+import os
+
+from yarl import URL
 
 from ..lib import utils, NETWORK_ERRORS
 from ..lib.externaldata import (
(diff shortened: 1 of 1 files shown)


wjt (approved):

src/checkers/jsonchecker.py
```diff
@@ -68,10 +71,18 @@ class JSONChecker(HTMLChecker):
     }
     SUPPORTED_DATA_CLASSES = [ExternalData, ExternalGitRepo]
 
-    async def _get_json(self, url: str) -> bytes:
+    async def _get_json(self, url: t.Union[str, URL]) -> bytes:
         log.debug("Get JSON from %s", url)
+        url = URL(url)
+
+        headers = {}
+        if url.host == "api.github.com":
```
gasinvein, line 79: Here we rely on that yarl reliably parses URLs and can't be tricked into thinking that the host is different from what it really is. Otherwise, we can send our token to a malicious host, since the URL is provided by user.

@barthalion If you consider that this check is not enough (or you're paranoid like me), we can use different tokens for querying GitHub APIs here and opening/merging PRs in main.py.

barthalion: It's fine by me the way it is now, thanks.

PR #241, opened 2021-10-06
Fix name of anitya checker test

wjt:


gasinvein:
Ah right, spelling "Anitya" still takes me some hesitation.

PR #242, opened 2021-10-06
anityachecker: Handle no matching versions case

### LINKED ISSUE #240: versions filter: failure when relational operation generate an empty list to target a future release
The scenario is that the module is need to be built from a specific commit, but soon as there's a new release then f-e-d-c should update the source.  

The problem is that the Anitya checker fails with the error `IndexError: list index out of range
` when adding a versions filter with a greater than operator, using the current last release as an operand.

I'm using an updated `org.flathub.flatpak-external-data-checker` app from Flathub.

Example module:
```
name: gnuradio-iqbal
buildsystem: cmake
sources:
  - type: git
    url: https://github.com/osmocom/gr-iqbal.git
    #tag: v0.38.2
    commit: fbee239a6fb36dd2fb564f6e6a0d393c4bc844db
    x-checker-data:
      type: anitya
      project-id: 14456
      stable-only: true
      versions:
        ">": 0.38.2
      tag-template: v$version
```

Log output:
```
INFO    src.checker: Started check [1/1] git gr-iqbal.git (from gnuradio-iqbal.yaml)
    main()
  File "/app/lib/fedc/src/main.py", line 395, in main
    return future.result()
  File "/app/lib/fedc/src/main.py", line 356, in run_with_args
  File "/app/lib/fedc/src/checker.py", line 341, in check
  File "/app/lib/fedc/src/checker.py", line 279, in _check_data
    await checker.check(data)
  File "/app/lib/fedc/src/checkers/anityachecker.py", line 58, in check
    latest_version = versions[0]
IndexError: list index out of range
```

gasinvein:
Fixes #240

### DIFF
--- a/src/checkers/anityachecker.py
+++ b/src/checkers/anityachecker.py
@@ -10,6 +10,7 @@ from ..lib.externaldata import (
     ExternalGitRef,
 )
 from ..lib.utils import filter_versions
+from ..lib.errors import CheckerQueryError
 from .htmlchecker import HTMLChecker
 
 log = logging.getLogger(__name__)
@@ -55,7 +56,10 @@ class AnityaChecker(HTMLChecker):
                 versions = result["versions"]
             if constraints:
                 versions = filter_versions(versions, constraints, sort=False)
-            latest_version = versions[0]
+            try:
+                latest_version = versions[0]
+            except IndexError as err:
+                raise CheckerQueryError("Can't find matching version") from err
         else:
             latest_version = result["latest_version"]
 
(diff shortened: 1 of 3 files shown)


src/checkers/anityachecker.py
```diff
@@ -55,7 +56,10 @@ async def check(self, external_data: ExternalBase):
                 versions = result["versions"]
             if constraints:
                 versions = filter_versions(versions, constraints, sort=False)
-            latest_version = versions[0]
+            try:
+                latest_version = versions[0]
+            except IndexError as err:
+                raise CheckerQueryError("Can't find matching version") from err
```
wjt, line 62: Might be nice to give more details about why not:

```suggestion
                raise CheckerQueryError(f"Can't find matching {'stable ' if stable_only else ''}version{ f' satisfying {constraints}' if constraints else ''}") from err
```

but having written that, it's really hard to read, and I guess it will be obvious by looking at the manifest.

wjt (approved):

src/checkers/anityachecker.py
gasinvein: >I guess it will be obvious by looking at the manifest.

Yeah. Also `constraints` will look messy when printed, since it's not just a `dict` but either `OrderedDict` or `CommentedMap`.

PR #243, opened 2021-10-10
refactor: Make source type a class attribute

gasinvein:
Use a single Python class to represent a single flatpak-builder source type.

This makes handling git type sources, which have a dedicated handler class, consistent with different file-like sources, which previously were handled by the same `ExternalData` class.
`.from_source()` class method now selects appropriate child class for handling given source.

We'll also need this change for some potential future changes:
- Handle type-specific source properties, e.g. `archive-type` for `archive` sources
- To implement source extracting, which is needed to run custom port-update hooks

### DIFF
--- a/src/lib/externaldata.py
+++ b/src/lib/externaldata.py
@@ -108,11 +108,7 @@ class ExternalBase(abc.ABC):
         if not source.get("url"):
             raise SourceUnsupported('Data is not external: no "url" property')
 
-        data_cls: t.Type[ExternalBase]
-        if data_type == cls.Type.GIT:
-            data_cls = ExternalGitRepo
-        else:
-            data_cls = ExternalData
+        data_cls = cls.data_classes()[data_type]
 
         return data_cls.from_source_impl(source_path, source)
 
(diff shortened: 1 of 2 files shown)


src/lib/externaldata.py
```diff
@@ -144,6 +140,15 @@ def set_new_version(
                 self.state = self.State.BROKEN
             self.new_version = new_version
 
+    @classmethod
+    def data_classes(cls) -> t.Set[t.Type[ExternalBase]]:
```
wjt: Shouldn't this return a dict from cls.type to cls? Then the usage above would be clearer.

wjt (changes requested):

src/lib/externaldata.py
wjt: If you were feeling really fancy you could (I think) use a metaclass to build the dictionary as the classes are defined.

gasinvein: >use a metaclass to build the dictionary

Nah, I don't think we want to go this far. I wanted to take inheritance into account (e.g. `ExternalData.from_source()` should raise and error if got a git type source), which I believe will be significantly more complex to implement using metaclass.

gasinvein: >Shouldn't this return a dict from cls.type to cls? Then the usage above would be clearer.

Yeah, done.

PR #259, opened 2021-12-10
Keep original base branch for pull requests

### LINKED ISSUE #258: Workflow opens PR at default branch instead of original branch
I wanted EasyEffects to be able to update nightly in flathub-beta without hassle. Since the hosted f-e-d-c only "targets" the default branch (usually master) in Flathub, it was suggested to reuse [this action](https://github.com/gasinvein/org.winehq.Wine/blob/branch/staging/.github/workflows/update.yml) within EasyEffects's Flathub repo instead, [like this](https://github.com/flathub/com.github.wwmm.easyeffects/blob/master/.github/workflows/update.yml).

However, despite triggering the workflow from [the beta branch](https://user-images.githubusercontent.com/38842733/145554435-49810aaf-231a-4f28-bce5-12d443371e50.png), the resulting PR [targeted master](https://github.com/flathub/com.github.wwmm.easyeffects/pull/41). It's possible to change the target PR branch after creation via GitHub's GUI, but it might introduce conflicts and is an extra step.

Ideally f-e-d-c could target the chosen branch for the PR instead of the default branch in the repo.

gasinvein:
Remember the initial branch we started at, propagate it to the `open_pr()` function and use as the base branch for opening pull request.

Fixes #258

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -117,8 +117,11 @@ def check_call(args):
     subprocess.check_call(args)
 
 
-def commit_changes(changes):
+def commit_changes(
+    changes: t.List[str],
+) -> t.Tuple[str, t.Optional[str], str, t.Optional[str]]:
     log.info("Committing updates")
+    body: t.Optional[str]
     if len(changes) > 1:
         subject = "Update {} modules".format(len(changes))
         body = "\n".join(changes)
@@ -127,6 +130,14 @@ def commit_changes(changes):
         subject = changes[0]
         body = None
         message = subject
+
+    # Remember the base branch
+    base_branch: t.Optional[str]
+    base_branch = subprocess.check_output(
+        ["git", "branch", "--show-current"], text=True
+    ).strip()
+    if not base_branch:
+        base_branch = None
 
     # Moved to detached HEAD
     check_call(["git", "checkout", "HEAD@{0}"])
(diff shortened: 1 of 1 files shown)


src/main.py
```diff
@@ -268,6 +282,12 @@ def open_pr(
 
     check_call(["git", "push", "-u", remote_url, branch])
 
+    log.info(
+        "Creting pull request in %s from head `%s` to base `%s`",
```
```suggestion
        "Creating pull request in %s from head `%s` to base `%s`",
```

wjt (approved):
LGTM with one typo fix and one suggestion for the future

src/main.py
```diff
@@ -117,8 +117,11 @@ def check_call(args):
     subprocess.check_call(args)
 
 
-def commit_changes(changes):
+def commit_changes(
+    changes: t.List[str],
+) -> t.Tuple[str, t.Optional[str], str, t.Optional[str]]:
```
wjt, line 122: At the point where a function returns a four-tuple, one might consider returning a namedtuple or a frozen dataclass.

gasinvein: Makes sense. Although I'd prefer to do this in a subsequent PR, since otherwise the change-set would be too big and harder to review.

src/main.py
gasinvein: Fixed, thanks.

src/main.py
wjt: Indeed

PR #261, opened 2021-12-22
Preserve trailing newline in json

### LINKED ISSUE #209: Keep trailing newline in JSON
We can't [keep](https://github.com/flathub/flatpak-external-data-checker/issues/192) most of the JSON formatting, but we could at least preserve trailing newline in the manifest, i.e. add it manually if it was there before we dump updated manifest.

cpba:
Preserve the trailing newline in json manifests by adding it
back if it was present before dumping the updated manifest.

Fixes #209

### DIFF
--- a/src/lib/utils.py
+++ b/src/lib/utils.py
@@ -68,6 +68,18 @@ def _extract_timestamp(info):
     raise CheckerRemoteError(f"Cannot parse date/time: {date_str}")
 
 
+def _check_newline(fp):
+    original_position = fp.tell()
+    fp.seek(0, os.SEEK_END)
+    fp.seek(fp.tell() - 1, os.SEEK_SET)
+    last_char = fp.read()
+    fp.seek(original_position, os.SEEK_SET)
+    if last_char == "\n":
+        return True
+    else:
+        return False
+
+
 def strip_query(url):
     """Sanitizes the query string from the given URL, if any. Parameters whose
     names begin with an underscore are assumed to be tracking identifiers and
@@ -425,13 +437,18 @@ def dump_manifest(contents, manifest_path):
     """Writes back 'contents' to 'manifest_path'.
 
     For YAML, we make a best-effort attempt to preserve
-    formatting; for JSON, we use the canonical 4-space indentation."""
+    formatting; for JSON, we use the canonical 4-space indentation,
+    but add a trailing newline if originally present."""
     _, ext = os.path.splitext(manifest_path)
+    with open(manifest_path, "r") as fp:
+        newline = _check_newline(fp)
     with open(manifest_path, "w", encoding="utf-8") as fp:
         if ext in (".yaml", ".yml"):
             _yaml.dump(contents, fp)
         else:
             json.dump(obj=contents, fp=fp, indent=4)
+            if newline:
+                fp.write("\n")
 
 
 def init_logging(level=logging.DEBUG):
--- a/tests/test_newlinepreservation.py
+++ b/tests/test_newlinepreservation.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python3
+# Copyright (C) 2021 Carles Pastor Badosa
+#
+# Authors:
+#       Carles Pastor Badosa <cpbadosa@gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+import unittest
+import os
+import tempfile
+
+from src.lib.utils import (
+    _check_newline,
+    dump_manifest,
+    read_manifest,
+)
+
+
+MANIFEST_WITH_NEWLINE = """{
+    "ends in newline": true
+}
+"""
+
+MANIFEST_NO_NEWLINE = """{
+    "ends in newline": false
+}"""
+
+
+class TestNewlinePreservation(unittest.TestCase):
+    def test_newline(self):
+        with tempfile.TemporaryDirectory() as d:
+            fp = os.path.join(d, "trailingnewline.json")
+            with open(fp, "w") as f:
+                f.write(MANIFEST_WITH_NEWLINE)
+            with open(fp, "r") as f:
+                self.assertTrue(_check_newline(f))
+            manifest = read_manifest(fp)
+            dump_manifest(manifest, fp)
+            with open(fp, "r") as f:
+                self.assertTrue(_check_newline(f))
+
+    def test_no_newline(self):
+        with tempfile.TemporaryDirectory() as d:
+            fp = os.path.join(d, "notrailingnewline.json")
+            with open(fp, "w") as f:
+                f.write(MANIFEST_NO_NEWLINE)
+            with open(fp, "r") as f:
+                self.assertFalse(_check_newline(f))
+            manifest = read_manifest(fp)
+            dump_manifest(manifest, fp)
+            with open(fp, "r") as f:
+                self.assertFalse(_check_newline(f))
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -1,0 +1,4 @@
+[mypy]
+
+[mypy-ruamel.yaml.*]
+ignore_errors = True
--- a/tests/test_checker.py
+++ b/tests/test_checker.py
@@ -392,7 +392,8 @@ size: {UpdateEverythingChecker.SIZE}
         "sha256": "{UpdateEverythingChecker.CHECKSUM}",
         "size": {UpdateEverythingChecker.SIZE}
     }}
-]""".lstrip()
+]
+""".lstrip()
         await self._test_update(
             filename,
             contents,


cpba:
Running the unit tests gives me back:
`FAILED (failures=1, expected failures=3)`

But I don't quite understand what I broke.

src/lib/utils.py
```diff
@@ -425,13 +433,17 @@ def dump_manifest(contents, manifest_path):
     """Writes back 'contents' to 'manifest_path'.
 
     For YAML, we make a best-effort attempt to preserve
-    formatting; for JSON, we use the canonical 4-space indentation."""
+    formatting; for JSON, we use the canonical 4-space indentation,
+    but add a trailing newline if originally present."""
     _, ext = os.path.splitext(manifest_path)
+    newline = _check_newline(manifest_path)
     with open(manifest_path, "w", encoding="utf-8") as fp:
```
gasinvein: It would be a bit better if we could avoid opening the output file twice. Maybe open it in read-write mode here, and pass `fp` to `_check_newline`? Just make sure to `seek` the file to proper positions when using read-write mode.

src/lib/utils.py
```diff
@@ -68,6 +68,14 @@ def _extract_timestamp(info):
     raise CheckerRemoteError(f"Cannot parse date/time: {date_str}")
 
 
+def _check_newline(manifest_path):
+    with open(manifest_path, "r") as f:
+        if f.read()[-1] == "\n":
```
gasinvein: This could be done in a less expensive way by seeking the file to the end and reading only the last char.

gasinvein (commented):
Here a some performance-related suggestions. 

Also please add a test case for this.

src/lib/utils.py
cpba: Sounds good, I'll change it and push a new commit.

cpba:
I believe the unit test here fails because it expects to get back a file without the trailing newline:
https://github.com/flathub/flatpak-external-data-checker/blob/master/tests/test_checker.py#L395

Should I modify it too?

gasinvein:
> I believe the unit test here fails because it expects to get back a file without the trailing newline: https://github.com/flathub/flatpak-external-data-checker/blob/master/tests/test_checker.py#L395
> 
> Should I modify it too?

Yeah, that should do.

cpba:
I forgot to run black again after the changes

src/lib/utils.py
```diff
@@ -68,6 +68,15 @@ def _extract_timestamp(info):
     raise CheckerRemoteError(f"Cannot parse date/time: {date_str}")
 
 
+def _check_newline(fp):
+    fp.seek(0, 2)
+    fp.seek(fp.tell()-1,0)
```
gasinvein: Please use named constants instead of numeric values, e.g. `os.SEEK_END` instead of `2`.
Also I guess this can be done in one action with negative offset instead of two, i.e. `fp.seek(-1, os.SEEK_END)`?

src/lib/utils.py
```diff
@@ -425,13 +434,19 @@ def dump_manifest(contents, manifest_path):
...
+    formatting; for JSON, we use the canonical 4-space indentation,
+    but add a trailing newline if originally present."""
     _, ext = os.path.splitext(manifest_path)
-    with open(manifest_path, "w", encoding="utf-8") as fp:
+    with open(manifest_path, "r+", encoding="utf-8") as fp:
         if ext in (".yaml", ".yml"):
             _yaml.dump(contents, fp)
         else:
+            newline = _check_newline(fp)
+            fp.seek(0, 0)
```
gasinvein: Nitpick: remember and return to the original position in `_check_newline` instead resetting it outside.

tests/test_util.py
```diff
@@ -222,5 +223,14 @@ async def test_wrong_content_type(self):
             )
 
 
+class TestCheckNewline(unittest.TestCase):
+    def test_newline(self):
+        fp = open("tests/com.jetbrains.PhpStorm.json", "r+", encoding="utf-8")
+        self.assertEqual(_check_newline(fp), True)
```
gasinvein: Ideally, we should do a full "newline preservation" test, i.e. create a dummy `.json` file with newline, read it, dump it, check that the newline is there, and do the same for a file without trailing newline.

src/lib/utils.py
```diff
@@ -68,6 +68,15 @@ def _extract_timestamp(info):
     raise CheckerRemoteError(f"Cannot parse date/time: {date_str}")
 
 
+def _check_newline(fp):
+    fp.seek(0, 2)
+    fp.seek(fp.tell() - 1, 0)
```
gasinvein: Please use named constants instead of numeric values, e.g. `os.SEEK_END` instead of `2`.
Also I guess this can be done in one action with negative offset instead of two, i.e. `fp.seek(-1, os.SEEK_END)`?

cpba: > Please use named constants instead of numeric values, e.g. os.SEEK_END instead of 2

Will do.

> Also I guess this can be done in one action with negative offset instead of two, i.e. fp.seek(-1, os.SEEK_END)?

Technically yes, but it requires opening the file in binary mode and I was afraid of edge cases with different encodings. Should I change it?

src/lib/utils.py
cpba: I agree, that would be less ugly.

tests/test_util.py
cpba: If I'm reading this correctly, should I add new json files specifically for this test rather than reuse the ones I did?

And the test should also check whether the json file is correctly generated with the trailing newline, rather than relying only on the existing test at https://github.com/flathub/flatpak-external-data-checker/blob/master/tests/test_checker.py#L141?

src/lib/utils.py
gasinvein: Right. I'm not sure if `json.dump` will play nice with binary IO, so it's probably safer to leave it as is, then.

tests/test_util.py
gasinvein: > should I add new json files specifically for this test rather than reuse the ones I did?

Yes, although you can embed the files contents into the test code instead of storing as real files in git.

>And the test should also check whether the json file is correctly generated with the trailing newline

Not necessary, but probably it would be even better.

So the test flow will look something like this:
1. Check if `_check_newline` works properly on a file with known contents
2. Read and dump again the file
3. Use `_check_newline` to check if trailing newline was preserved

cpba:
Alright, I think it's looking more presentable now.

src/lib/utils.py
```diff
@@ -425,13 +436,18 @@ def dump_manifest(contents, manifest_path):
...
-    with open(manifest_path, "w", encoding="utf-8") as fp:
+    with open(manifest_path, "r+", encoding="utf-8") as fp:
         if ext in (".yaml", ".yml"):
             _yaml.dump(contents, fp)
         else:
+            newline = _check_newline(fp)
             json.dump(obj=contents, fp=fp, indent=4)
+            if newline:
+                fp.write("\n")
+        fp.truncate()
```
gasinvein: I'm not sure I understand why this is needed.

src/lib/utils.py
```diff
@@ -68,6 +68,17 @@ def _extract_timestamp(info):
     raise CheckerRemoteError(f"Cannot parse date/time: {date_str}")
 
 
+def _check_newline(fp):
+    fp.seek(0, os.SEEK_END)
+    fp.seek(fp.tell() - 1, os.SEEK_SET)
+    last_char = fp.read()
+    fp.seek(0, os.SEEK_SET)
```
gasinvein: Nitpick: It would nicer to remember the actual initial position and reset to it after we're done, instead of resetting to the beginning.

tests/test_newlinepreservation.py
```diff
@@ -0,0 +1,149 @@
...
+
+from src.lib.utils import (
+    _check_newline,
+    dump_manifest,
+    read_manifest,
+)
+
+
+MANIFEST_WITH_NEWLINE = """{
+  "app-id": "com.gitlab.coringao.cavestory-nx",
```
gasinvein: We don't need an actual flatpak manifest here to test dict dumping. Any json would do, even something like `{"one": 1}`.

src/lib/utils.py
cpba: I thought there might potentially be cases where the new file is shorter than the original, since we're using `r+` mode we could end up with garbage at the end of the file if we don't truncate.

tests/test_newlinepreservation.py
cpba: It's one of mine, I just had it lying around and threw it in for realism. Would it be better to replace it with a minimal one liner?

src/lib/utils.py
cpba: Yeah that'd be safer.

src/lib/utils.py
gasinvein: Ah, indeed, good catch. I didn't think about it when suggested to use read-write mode. Now opening opening the file twice - in read-only and write-only modes - seems like a better idea :-/

tests/test_newlinepreservation.py
gasinvein: I think it will, yeah. No need to add more text than necessary.

src/lib/utils.py
cpba: Do you want me to go back to the original implementation?

gasinvein: Yeah, if you don't mind, sorry for the confusion. Just don't change `_check_newline` argument back to a file path - instead, open the file for reading and detecting trailing newline here, in `dump_manifest` (maybe better outside this `with open()` block).

cpba: So if I understand this right, you want me to keep `_check_newline` as it currently is but just change `dump_manifest` back to `w` mode and open it for reading beforehand? I'll send a new commit and see if that's what you want.

gasinvein: Yep, something like this.

gasinvein (approved):

gasinvein:
@cpba Thanks for your contribution. Don't mind if I squash your commits, since most of them redid previous ones?

cpba:
No, I don't mind if you squash them, otherwise I'd have to go clean up my
commit messages :)

gasinvein:
Ouch. Looks like Mypy was updated and now complains about ruamel.yaml, failing our type check test. I guess it's unrelated to this PR.

PR #263, opened 2022-01-06
Respect some editorconfig settings

### LINKED ISSUE #256: Tabs in Flatpak manifests are converted to spaces
As can be seen in this diff, tabs are for some reason converted to spaces by f-e-d-c when it issues an upgrade: https://github.com/flathub/com.jpexs.decompiler.flash/pull/9/files

I prefer to use tabs over spaces for various reasons. Is this a deliberate feature or something that could be fixed?

### LINKED ISSUE #262: Override formatting default with editorconfig when present
Editorconfig file is a standard way to declare eg indentation (what to use, how wide). Currently there is no way to get f-e-d-c abide to formatting in manifest files. If the tool respected editorconfig, this would alleviate the issue and also potentially encourage app maintainers to use editorconfig files more widely.

gasinvein:
We can use the editorconfig library to get settings for a manifest we're about to dump, and, depending on it, ajust indentation style and size (in case of spaces), and determine if we should add trailing newline.

Fixes #262 
Closes #256
Adds another way to address #209

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,12 @@ from the tool can be forcefully enabled by setting
 `automerge-flathubbot-prs` to `true` in `flathub.json`,
 or it can be completely disabled by setting `automerge-flathubbot-prs` to `false`.
 
+### Formatting manifests
+
+The tool respects some of [EditorConfig](https://editorconfig.org/) settings.
+If your preferred manifest formatting is different from the default (4-space indentation for JSON),
+create (and commit) a `.editorconfig` file near the manifest.
+
 ## Changes to Flatpak manifests
 
 For simple checks to see if a URL is broken, no changes are needed.  However,
--- a/README.md
+++ b/README.md
@@ -89,9 +89,21 @@ or it can be completely disabled by setting `automerge-flathubbot-prs` to `false
 
 ### Formatting manifests
 
-The tool respects some of [EditorConfig](https://editorconfig.org/) settings.
-If your preferred manifest formatting is different from the default (4-space indentation for JSON),
-create (and commit) a `.editorconfig` file near the manifest.
+When writing back JSON files, this tool defaults to four-space indentation, preserving or omitting a trailing newline based on the source file. If you prefer a different formatting, create and commit an `.editorconfig` file describing your preferred formatting for `.json` files. At present, this tool respects a subset of [EditorConfig](https://editorconfig.org/) settings:
+
+```ini
+root = true
+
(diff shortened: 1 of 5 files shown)


src/lib/utils.py
```diff
@@ -418,35 +419,58 @@ def read_json_manifest(manifest_path):
...
 
     For YAML, we make a best-effort attempt to preserve
     formatting; for JSON, we use the canonical 4-space indentation,
     but add a trailing newline if originally present."""
-    _, ext = os.path.splitext(manifest_path)
-    with open(manifest_path, "r") as fp:
-        newline = _check_newline(fp)
-    with open(manifest_path, "w", encoding="utf-8") as fp:
-        if ext in (".yaml", ".yml"):
+    if not isinstance(manifest_path, Path):
```
nanonyme: This is probably unnecessary check. You could just call Path(manifest_path) even if given Path to begin with.

gasinvein: Indeed, thanks.

gasinvein:
Added a minimal test case.

wjt:
I haven't had a chance to review the code but I think this would be worth a prominent note in the README (the current best documentation) since this would resolve quite a common complaint.

gasinvein:
@wjt Makes sense, added a note about the manifest formatting.

wjt (approved):
LGTM. I pushed a slightly more detailed README entry.

I do wonder whether there already exists a library which can write out formats like JSON & YAML and respect as much of EditorConfig as possible, rather than open-coding it like this. But I couldn't find anyting in a quick search

README.md
```diff
@@ -87,6 +87,24 @@ from the tool can be forcefully enabled by setting
...
+[*.json]
+indent_style = space
+# Only integer values are supported; ignored if indent_style=tab
+indent_size = 2
+insert_final_newline = true
+```
+
+Unfortunately, it is not feasible to preserve JSON-GLib's non-standard `/* */` syntax for comments. As an alternative, dictionary keys beginning with `//` are ignored by `flatpak-builder` and can be used for comments in many cases.
+
+For YAML files, this tool attempts to preserve existing formatting and comments automatically. `.editorconfig` is not used.
```
gasinvein, line 106: Flathub has a [YAML style guide](https://github.com/flathub/flathub/wiki/YAML-Style-Guide), which every app maintainer should adhere to anyway; probably it would've made sense to mention it here.

PR #264, opened 2022-01-07
README: Advertise version constraints in modules

martymichal:


README.md
```diff
@@ -266,11 +266,18 @@ and add a template for source download URL.
     "type": "anitya",
     "project-id": 6377,
     "stable-only": false,
+    "versions": {"<": "1.12.0"},
     "url-template": "https://github.com/flatpak/flatpak/releases/download/$version/flatpak-$version.tar.xz"
 }
 ```
 Set `stable-only` to `true` to retrieve latest stable version (as recognized by Anitya).
 
+You can also set version constraints in `versions` property (optional). It should
```
gasinvein: Ideally, we probably should put this paragraph somewhere once, and reference it from descriptions of all checkers that support the constraints, instead of repeating the same text in multiple places.

martymichal: Tried to unify the wording. Is the new version acceptable?

README.md
```diff
@@ -116,6 +116,14 @@ and automatically add it to releases list in metainfo. To specify which source
 is the app upstream source, set property `is-main-source` to `true` in the
 checker metadata for that source.
 
+The checkers can also support version constraining, making it possible to e.g.
```
gasinvein: Use a sub-section here (with header e.g. `### Version constraining`), then it can be referenced from checkers descriptions like `The [versions](#version-constraining) property is supported`.

martymichal: Fixed.

gasinvein:
Thanks!

PR #265, opened 2022-01-07
lib/utils: Catch type errors on loose version comparison

martymichal:
Some software[0] does not strictly use numbers or semantic versioning
for labeling tags/releases in git. Trying to filter out these versions
would lead to TypeError due to comparison between int and str (str being
the non-compliant version string). Catch this error and due to no
reliable of checking the version, mark it as a non-match.

[0] https://release-monitoring.org/project/13289/

### DIFF
--- a/src/lib/utils.py
+++ b/src/lib/utils.py
@@ -156,7 +156,14 @@ def filter_versions(
             try:
                 match = oper(StrictVersion(version_str), StrictVersion(version_limit))
             except ValueError:
-                match = oper(LooseVersion(version_str), LooseVersion(version_limit))
+                try:
+                    match = oper(LooseVersion(version_str), LooseVersion(version_limit))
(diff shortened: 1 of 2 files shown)


src/lib/utils.py
```diff
@@ -166,7 +166,10 @@ def filter_versions(
             try:
                 match = oper(StrictVersion(version_str), StrictVersion(version_limit))
             except ValueError:
-                match = oper(LooseVersion(version_str), LooseVersion(version_limit))
+                try:
+                    match = oper(LooseVersion(version_str), LooseVersion(version_limit))
+                except TypeError:
```
gasinvein: Please capture and debug-log the exception.

martymichal: Added the debug log.

gasinvein:
Thanks!

Issue #222, opened 2021-08-24
Support updating release notes in metainfo


vchernin:
Today f-e-d-c can update the version number and release date in the metainfo/appdata file. I wonder if this could be extended to support adding the changelog details for each update to metainfo/appdata. This would require finding ways to parse an upstream changelog/news file. [Relevant link](https://blog.tenstral.net/2020/03/maintain-release-info-easily-in-metainfo-files.html)

This would ideally improve the situation for many apps on Flathub (eg Chromium) that have a version history like this:
<img src="https://user-images.githubusercontent.com/38842733/130543297-90cb0c56-3aaf-45ae-b6c5-1c5bec3700e6.png" alt="gnome software version info" width="400"/>

wjt:
Yes, that would be great in theory. I fear it may be difficult to build something generic that can parse release information that is intended for end-users rather than developers.

Using Chromium as an example, is there an existing source that could work for this? The best I have found is "Stable Channel Update" posts on https://chromereleases.googleblog.com/ 

The alternative is for the human who merges each PR to add the news by hand.

gasinvein:
Hmm, I as well doubt this could be possible in a generic way. Yet there are sources that provide release info we can grab, like release notes on GitHub. Maybe we should add a GitHubChecker?

gasinvein:
Also, when/if we get either a working sandbox in github actions or a per-app f-e-d-c action, we can allow running custom scripts to retrieve release info during update by f-e-d-c.

vchernin:
See also: https://github.com/ximion/appstream/issues/354

nanonyme:
It would really be best approach to get as many projects as possible ship their appdata files that contain release notes. Core problem of this need to begin with is data that is supposed to come from upstream is getting patched together on Flathub side.

vchernin:
I think perhaps more automated ways to maintain appdata release notes would help upstreams, https://blog.tenstral.net/2020/03/maintain-release-info-easily-in-metainfo-files.html helps but there are so many ways projects write release notes.

nanonyme:
Does it btw support changelog? I noticed mosh actually has NEWS that is abandoned (?) and doesn't have newer releases, newer hings live in a file like https://github.com/mobile-shell/mosh/blob/master/ChangeLog.

hadess:
[...]

PR #267, opened 2022-01-10
appdata: Rewrite appdata updater using lxml

gasinvein:
lxml can preserve most of the XML formatting OOTB, including comments, so we need much less code to handle indentation for newely added nodes.
This patch also separates XML structure changes logic from formatting.
Trailing newlines currently aren't preserved (they're always added instead). No other behaviour changes expected.

This is the first step needed to allow re-indenting appdata XMLs (which, in turn, will allow us to support and editorconfig for appdata, and ultimately implement #222).

### DIFF
--- a/src/lib/appdata.py
+++ b/src/lib/appdata.py
@@ -23,123 +23,67 @@ preserving as much formatting as is feasible and inserting that element if it is
 missing.
 """
 
-from io import StringIO
-from defusedxml.sax import make_parser
-from xml.sax.handler import property_lexical_handler
-from xml.sax.saxutils import XMLFilterBase, XMLGenerator
+import io
+import typing as t
+
+import lxml.etree as ET
 
 
-class AddVersionFilter(XMLFilterBase):
-    def __init__(self, version, date, parent=None):
-        super().__init__(parent)
-
-        self._version = version
-        self._date = date
-        self._context = []
-        self._emitted_release = False
-        self._releases_padding = ""
-
-    @property
-    def outside_root_element(self):
-        return not self._context
-
-    @property
-    def _in_releases(self):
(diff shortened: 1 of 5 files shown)


src/lib/appdata.py
```diff
@@ -23,123 +23,67 @@
...
+
+    release = ET.Element("release", version=version, date=date)
+    releases.insert(0, release)
+    _fill_padding(release)
+
+    tree.write(
+        dst,
+        # XXX: lxml uses single quotes for doctype line if generated with
+        # xml_declaration=True,
+        doctype='<?xml version="1.0" encoding="UTF-8"?>',
```
wjt, line 78: Re. the XXX: so is that bad? Is the issue here that the prevailing style for Flatpak apps is to use double-quotes?

wjt (approved):
Looks great. I'm surprised it's so simple and I wonder why I didn't do this in 2019.

src/lib/appdata.py
gasinvein: I usually use `XXX` comments for "we do this weird thing because I don't know a better way to achieve that". Is it misplaced?
>Is the issue here that the prevailing style for Flatpak apps is to use double-quotes?

Yep, almost everyone seems to use double quotes.

wjt: :shrug: whatever works – I just wasn't sure from the comment whether this was something to fix later, tolerate, or something  else.

gasinvein: I'd say "fix when/if becomes possible, tolerate until then".

gasinvein:
I've forgot to adapt `checker.py` for lxml usage. Fixed.

wjt (approved):

PR #270, opened 2022-01-25
Electron Auto Update checker

### LINKED ISSUE #268: YAML checker
need to construct url using a yml file provided by upstream

gasinvein:
Electron has an [auto-update mechanism](https://www.electron.build/auto-update.html), which relies on remotely stored metadata file (e.g. `latest-linux.yml`). We can use this metadata, too.
This is the first checker to rely on #208, since the electron-updater metadata seem to usually contain only sha512 checksums.
Closes #268

### DIFF
--- a/src/checkers/electronchecker.py
+++ b/src/checkers/electronchecker.py
@@ -1,0 +1,83 @@
+import logging
+import base64
+from datetime import datetime
+import typing as t
+
+from yarl import URL
+import ruamel.yaml
+
+from ..lib import NETWORK_ERRORS
+from ..lib.externaldata import (
+    Checker,
+    ExternalBase,
+    ExternalData,
+    ExternalFile,
+)
+from ..lib.errors import CheckerQueryError
+from ..lib.checksums import MultiDigest
+from .jsonchecker import parse_timestamp
+
+yaml = ruamel.yaml.YAML(typ="safe")
+log = logging.getLogger(__name__)
+
+
+class ElectronChecker(Checker):
+    CHECKER_DATA_TYPE = "electron-updater"
+    CHECKER_DATA_SCHEMA = {
+        "type": "object",
+        "properties": {
+            "url": {"type": "string", "format": "uri"},
+        },
+    }
+
+    @staticmethod
+    def _read_digests(obj: t.Dict) -> MultiDigest:
+        digests: t.Dict[str, str] = {}
+        for _k in MultiDigest._fields:  # pylint: disable=no-member
+            if _k in obj:
+                digests[_k] = base64.b64decode(obj[_k]).hex()
+        return MultiDigest(**digests)
+
+    async def check(self, external_data: ExternalBase):
+        assert isinstance(external_data, ExternalData)
+
+        if "url" in external_data.checker_data:
+            metadata_url = URL(external_data.checker_data["url"])
+        else:
+            metadata_url = URL(external_data.current_version.url).join(
+                URL("latest-linux.yml")
+            )
+
+        try:
+            async with self.session.get(metadata_url) as resp:
+                metadata = yaml.load(await resp.read())
+        except NETWORK_ERRORS as err:
+            raise CheckerQueryError from err
+
+        if "files" in metadata:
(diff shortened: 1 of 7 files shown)


src/checkers/electronchecker.py
```diff
@@ -0,0 +1,83 @@
...
+
+        try:
+            async with self.session.get(metadata_url) as resp:
+                metadata = yaml.load(await resp.read())
+        except NETWORK_ERRORS as err:
+            raise CheckerQueryError from err
+
+        if "files" in metadata:
+            # Modern metadata format
+            m_file = metadata["files"][0]
```
wjt, line 59: Is it always the first and only file in this array?

src/checkers/electronchecker.py
```diff
@@ -0,0 +1,83 @@
...
+            file_size = int(m_file["size"])
+            checksum = self._read_digests(m_file)
+        else:
+            # Old electron-updater 1.x metadata format; no size property
+            file_url = metadata_url.join(URL(metadata["path"]))
+            file_size = None
+            checksum = self._read_digests(metadata)
+
+        timestamp: t.Optional[datetime]
+        if isinstance(metadata["releaseDate"], datetime):
```
wjt, line 70: In what case will it have been transparently converted to a datetime by ruamel?

src/lib/externaldata.py
```diff
@@ -454,3 +454,51 @@ async def validate_checker_data(self, external_data: ExternalBase):
...
+        Set the `new_version` for `source`, ensuring common digest types are set.
+        """
+        if (
+            isinstance(source, ExternalData) and isinstance(new_version, ExternalFile)
+        ) and not (
+            source.current_version.checksum.digests & new_version.checksum.digests
+        ):
+            log.warning(
+                "Source %s: %s didn't get a suitable checksum type for this source; "
+                "consider replacing %s with %s in the manifest",
```
wjt: I believe but could be wrong that extra-data sources can only have one checkum type, sha256 iirc. I think this advice is only applicable in the case where a normal source is being checked?

wjt (approved):

src/checkers/electronchecker.py
gasinvein: I was surprised to find out that [timestamp](https://yaml.org/type/timestamp.html) is actually a valid type recognized at YAML level. `ruamel.yaml` seems to always convert it to python `datetime`.

src/checkers/electronchecker.py
gasinvein: Honestly, I'm not sure. [Documentation](https://www.electron.build/auto-update.html) doesn't cover it (or I didn't find it). But from the [electron-updater](https://github.com/electron-userland/electron-builder/tree/master/packages/electron-updater) code, it seems like currently the `file` list always contains only one entry.

src/lib/externaldata.py
gasinvein: Indeed, extra-data only supports sha256, so this advice may be misleading. Should we change the advice to mention the extra-data limitation, or just not give the advice in case of extra-data?

wjt: I think it would be better to not give the advice at all for extra-data sources.

src/checkers/electronchecker.py
wjt: Surprising indeed!

src/lib/externaldata.py
```diff
@@ -454,3 +454,50 @@ async def validate_checker_data(self, external_data: ExternalBase):
...
+        """
+        Set the `new_version` for `source`, ensuring common digest types are set.
+        """
+        if (
+            isinstance(source, ExternalData) and isinstance(new_version, ExternalFile)
+        ) and not (
+            source.current_version.checksum.digests & new_version.checksum.digests
+        ):
+            log.warning(
+                "Source %s: %s didn't get a %s digest; available digests were %s",
```
gasinvein, line 492: @wjt Is the wording better now?

wjt: LGTM

PR #275, opened 2022-02-10
Fix run-in-container wrapper script and associated instructions

vchernin:
The container image was adapted for local use in 064fdeaa85e00bf72832d1991a571b4e009137ae, which means the default entrypoint is f-e-d-c itself.
However, test suites from CONTRIBUTING.md don't run, since they need to be run from python3 normally. So fix by setting entrypoint to python3 in the run-in-container wrapper script.

Black was removed since it is run by CI and no longer present in the image, and should be easy to install in a Toolbox or elsewhere.

### DIFF
--- a/run-in-container.sh
+++ b/run-in-container.sh
@@ -6,6 +6,13 @@ GIT_USER_EMAIL=$(git config user.email)
 
 CWD=$(pwd)
 
+PYTHON_ENTRYPOINT=""
+
+if [ "$1" == "python3" ]; then
+    PYTHON_ENTRYPOINT="--entrypoint=python3"
+    shift # don't pass first argument since setting entrypoint
+fi
+
 podman run --rm --privileged \
     -v $HOME:$HOME:rslave \
     -v $CWD:$CWD:rslave \
@@ -14,5 +21,6 @@ podman run --rm --privileged \
     -e GIT_COMMITTER_NAME="$GIT_USER_NAME" \
     -e GIT_AUTHOR_EMAIL="$GIT_USER_EMAIL" \
     -e GIT_COMMITTER_EMAIL="$GIT_USER_EMAIL" \
+    $PYTHON_ENTRYPOINT \
     -it ghcr.io/flathub/flatpak-external-data-checker \
     $*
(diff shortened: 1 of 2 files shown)


run-in-container.sh
```diff
@@ -14,5 +14,6 @@ podman run --rm --privileged \
     -e GIT_COMMITTER_NAME="$GIT_USER_NAME" \
     -e GIT_AUTHOR_EMAIL="$GIT_USER_EMAIL" \
     -e GIT_COMMITTER_EMAIL="$GIT_USER_EMAIL" \
+    --entrypoint=python3 \
```
gasinvein: Doesn't this mean `run-in-container.sh` won't run f-e-d-c anymore, unless its script is manually passed on cli?

vchernin: You're right, I somehow missed that `run-in-container.sh` was also used for just running f-e-d-c. I've changed it so it selects `python3` as the entrypoint if `python3` happens to be the first argument. It works for both running f-e-d-c normally and running the `CONTRIBUTING.md` tests.

vchernin:
I have been using this for my important source PR, it works as intended, even if a bit hacky

gasinvein:
Honestly, I'm not sure why we need this wrapper script at all. Running fedc with `podman` directly seems easy enough to me.
Lets ask @wjt opinion.

wjt:
I am not attached to this script.

gasinvein:
@wjt I suppose Endless doesn't need it on its CI anymore (IIRC it did)?

wjt:
No, all that is long gone.

wjt:
Stumbled back upon this. At this point I think the wrapper script is actively misleading and I guess we should kill it and update the documentation. But I did find its old usage useful for local development…

bbhtt:
The script is useful for local dev indeed particularly for tests. the current issue was fixed in https://github.com/flathub-infra/flatpak-external-data-checker/commit/27ec3101c2c346169be25d046899bc52b17bd458 and https://github.com/flathub-infra/flatpak-external-data-checker/commit/e67f6c3edb444ae84a5233671594536982f76991

PR #276, opened 2022-02-22
Add support for selectively updating manifests

### LINKED ISSUE #249: Support only opening PRs for certain sources
This was discussed on Matrix a bit ago, but I couldn't find an issue so I opened one.

It would be useful for maintainers to only get PRs when it is really needed. A manifest may include many instances of x-checker-data in their manifest. But this means often PRs will be opened for library updates and the Flatpak probably don't need to be urgently updated. 

So it would handy if f-e-d-c could be told to only create PRs under certain circumstances. One configuration could be to only open an update PR when the main source is updated.

I see at least two ways of approaching this:

1. Current x-checker-data works as it already does. Add an option to opt-out particular sources of being critical. This means update PRs will not be made for that source unless the main or a critical source gets an update.

2. x-checker-data defaults are changed. Update PRs will only be opened if the main source gets an update. Override this assumption by marking a source as critical, which means a PR is opened as soon as a source gets an update.

From a security perspective it seems somewhat reasonable to update things as soon as possible. But that can easily become excessive. I am not sure what the sane default is here.

vchernin:
Fixes https://github.com/flathub/flatpak-external-data-checker/issues/249

I opted to add an optional property to mark sources as "important", `is-important: true`. If "important" sources are present in external data, only update the manifest if at least 1 important source received an update. This basically lets apps choose when they want PRs to open, e.g. many apps may only want a PR when at least their primary upstream app source is updated. 

Manifests without important sources will not be affected by this change. f-e-d-c will continue making manifest updates as before for all modules. In other words, this does not change any existing behaviour.

There are some (I think comprehensive) tests for this feature in `test_checker.py`.

I also tested with an actual GitHub repo (forked EasyEffects which happens to have convenient action to use):

> [When is-critical-source is set for EasyEffects itself, but only libsigc++ has an update available](https://github.com/vchernin/com.github.wwmm.easyeffects/runs/5294315473?check_suite_focus=true)

Doesn't make a PR/manifest update as shown by the logs as expected

> [When is-critical-source is set for both for EasyEffects and libsigc++, and libsigc++ has an update available](https://github.com/vchernin/com.github.wwmm.easyeffects/runs/5294368744?check_suite_focus=true)

Makes [PR as expected](https://github.com/vchernin/com.github.wwmm.easyeffects/pull/5)

### DIFF
--- a/src/lib/externaldata.py
+++ b/src/lib/externaldata.py
@@ -47,6 +47,7 @@ CHECKER_DATA_SCHEMA_COMMON = {
     "properties": {
         "type": {"type": "string"},
         "is-main-source": {"type": "boolean"},
+        "is-important": {"type": "boolean"},
         "arches": {
             "type": "array",
             "items": {"type": "string"},
@@ -275,6 +276,11 @@ class ExternalBase(BuilderSource):
 
             self.new_version = new_version
 
+    @property
+    def has_version_changed(self) -> bool:
+        """Detect if some external data had a version change"""
+        raise NotImplementedError
+
     def update(self):
         """If self.new_version is not None, writes back the necessary changes to the
         original element from the manifest."""
@@ -371,6 +377,21 @@ class ExternalData(ExternalBase):
                     self.source_path,
                 )
                 self.source.pop("size", None)
+
+    @property
+    def has_version_changed(self) -> bool:
+        update = self.new_version
+        return (
+            update is not None
+            and update.version is not None
+            and isinstance(update, ExternalFile)
+            and (
+                update.url != self.current_version.url
+                # TODO We can't reliably tell if the appimage version stayed the same
+                # without downloading it, so just assume it changed
+                or update.url.endswith(".AppImage")
+            )
+        )
 
 
 @dataclasses.dataclass
@@ -501,6 +522,16 @@ class ExternalGitRepo(ExternalBase):
             None,
         )
         return obj
+
+    @property
+    def has_version_changed(self) -> bool:
+        update = self.new_version
+        return (
+            update is not None
+            and update.version is not None
+            and isinstance(update, ExternalGitRef)
+            and update.tag != self.current_version.tag
+        )
 
     def update(self):
         if self.new_version is not None:
--- a/src/main.py
+++ b/src/main.py
@@ -392,6 +392,13 @@ def parse_cli_args(cli_args=None):
         type=int,
         default=manifest.MAX_MANIFEST_SIZE,
     )
+    parser.add_argument(
+        "--require-important-update",
+        help="Require an update to at least one source with is-important or is-main-source to save changes to the manifest. "
+        "If no instances of is-important or is-main-source are found, assume normal behaviour and always save changes to the manifest. "
+        "This is useful to avoid PRs generated to update a singular unimportant source.",
+        action="store_true",
+    )
 
     return parser.parse_args(cli_args)
 
@@ -405,6 +412,7 @@ async def run_with_args(args: argparse.Namespace) -> t.Tuple[int, int, bool]:
     options = manifest.CheckerOptions(
         allow_unsafe=args.unsafe,
         max_manifest_size=args.max_manifest_size,
+        require_important_update=args.require_important_update,
     )
 
     manifest_checker = manifest.ManifestChecker(args.manifest, options)
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -56,6 +56,7 @@ from .checkers import ALL_CHECKERS
 
 
 MAIN_SRC_PROP = "is-main-source"
+IMPORTANT_SRC_PROP = "is-important"
 MAX_MANIFEST_SIZE = 1024 * 100
 
 
@@ -78,6 +79,7 @@ def find_appdata_file(directory, appid):
 class CheckerOptions:
     allow_unsafe: bool = False
     max_manifest_size: int = MAX_MANIFEST_SIZE
+    require_important_update: bool = False
 
 
 class ManifestChecker:
@@ -495,27 +497,7 @@ class ManifestChecker:
 
         last_update = selected_data.new_version
 
-        version_changed = (
-            last_update is not None
-            and last_update.version is not None
-            and (
-                (
-                    isinstance(last_update, ExternalFile)
-                    and (
-                        last_update.url != selected_data.current_version.url
-                        # TODO We can't reliably tell if the appimage version stayed the same
-                        # without downloading it, so just assume it changed
-                        or last_update.url.endswith(".AppImage")
-                    )
-                )
-                or (
-                    isinstance(last_update, ExternalGitRef)
-                    and last_update.tag != selected_data.current_version.tag
-                )
-            )
-        )
-
-        if version_changed:
+        if selected_data.has_version_changed:
             log.info("Version changed, adding release to %s", appdata)
             if last_update.timestamp is None:
                 log.warning("Using current time in appdata release")
@@ -532,20 +514,46 @@ class ManifestChecker:
             log.debug("Version didn't change, not adding release")
 
     def update_manifests(self) -> t.List[str]:
-        """Updates references to external data in manifests."""
+        """Updates references to external data in manifests.
+        If require_important_update is True, only update the manifest
+        if at least one source with IMPORTANT_SRC_PROP or MAIN_SRC_PROP received an update."""
         # We want a list, without duplicates; Python provides an
         # insertion-order-preserving dictionary so we use that.
         changes: t.Dict[str, t.Any]
         changes = OrderedDict()
-        for path, datas in self._external_data.items():
-            self._update_manifest(path, datas, changes)
-        if changes:
-            try:
-                self._update_appdata()
-            except AppdataNotFound as err:
-                log.info("Not updating appdata: %s", err)
-            except AppdataError as err:
-                self._errors.append(err)
-                log.error(err)
+
+        found_important_update = None
+
+        if self.opts.require_important_update:
+
+            for data in self.get_external_data():
+                important = data.checker_data.get(IMPORTANT_SRC_PROP)
+                main = data.checker_data.get(MAIN_SRC_PROP)
+                if important or (main and important != False):
+                    log.debug("Found an important source: %s", data)
+
+                    found_important_update = data.has_version_changed
+
+                    if found_important_update:
+                        log.info(
+                            "Update found for important source: %s, updating manifest",
+                            data,
+                        )
+                        break
+
+        if found_important_update or found_important_update is None:
+            for path, datas in self._external_data.items():
+                self._update_manifest(path, datas, changes)
+            if changes:
+                try:
+                    self._update_appdata()
+                except AppdataNotFound as err:
+                    log.info("Not updating appdata: %s", err)
+                except AppdataError as err:
+                    self._errors.append(err)
+                    log.error(err)
+
+        else:
+            log.info("No important source had an update, not updating manifest")
 
         return list(changes)
--- a/tests/test_checker.py
+++ b/tests/test_checker.py
@@ -26,15 +26,18 @@ import tempfile
 import hashlib
 import base64
 from xml.dom import minidom
+import typing as t
 
 import aiohttp
 
 from src.lib.utils import init_logging
 from src.lib.externaldata import ExternalData
 from src.lib.checkers import Checker
+from src.checkers.gitchecker import GitChecker
+from src.lib.externaldata import ExternalGitRepo
 from src.lib.checksums import MultiDigest
 from src.lib.errors import CheckerFetchError
-from src.manifest import ManifestChecker
+from src import manifest
 
 TEST_MANIFEST = os.path.join(
     os.path.dirname(__file__),
@@ -84,30 +87,9 @@ class UpdateEverythingChecker(DummyChecker):
         )
 
 
-class TestExternalDataChecker(unittest.IsolatedAsyncioTestCase):
-    def setUp(self):
-        init_logging()
-
-    async def test_check_filtered(self):
-        # Use only the URLChecker which is fast so we don't have to wait a lot
-        # for this test; so save the real checkers for later
-        dummy_checker = ManifestChecker(TEST_MANIFEST)
-        dummy_checker._checkers = [DummyChecker]
-
-        ext_data = await dummy_checker.check()
-        ext_data_from_getter = dummy_checker.get_external_data()
-        self.assertEqual(len(ext_data), len(ext_data_from_getter))
-        self.assertEqual(ext_data, ext_data_from_getter)
-
-        self.assertEqual(len(ext_data), NUM_ALL_EXT_DATA)
-        ext_data = await dummy_checker.check(ExternalData.Type.EXTRA_DATA)
-        self.assertEqual(len(ext_data), NUM_EXTRA_DATA_IN_MANIFEST)
-
-        ext_data = await dummy_checker.check(ExternalData.Type.FILE)
-        self.assertEqual(len(ext_data), NUM_FILE_IN_MANIFEST)
-
-        ext_data = await dummy_checker.check(ExternalData.Type.ARCHIVE)
-        self.assertEqual(len(ext_data), NUM_ARCHIVE_IN_MANIFEST)
+class _TestWithInlineManifest(unittest.IsolatedAsyncioTestCase):
+    _DUMMY_CHECKER_CLS: t.Type[Checker]
+    maxDiff = None
 
     async def _test_update(
         self,
@@ -117,10 +99,11 @@ class TestExternalDataChecker(unittest.IsolatedAsyncioTestCase):
         expected_updates,
         expected_data_count=1,
         new_release=True,
+        require_important_update=False,
     ):
         with tempfile.TemporaryDirectory() as tmpdir:
-            manifest = os.path.join(tmpdir, filename)
-            with open(manifest, "w") as f:
+            manifest_path = os.path.join(tmpdir, filename)
+            with open(manifest_path, "w") as f:
                 f.write(contents)
 
             appdata = os.path.join(
@@ -130,16 +113,20 @@ class TestExternalDataChecker(unittest.IsolatedAsyncioTestCase):
             with open(appdata, "w") as f:
                 f.write("""<application></application>""")
 
-            checker = ManifestChecker(manifest)
+            options = manifest.CheckerOptions(
+                require_important_update=True,
+            )
+
+            checker = manifest.ManifestChecker(manifest_path, options)
             self.assertEqual(
                 len(checker.get_external_data()),
                 expected_data_count,
             )
-            checker._checkers = [UpdateEverythingChecker]
+            checker._checkers = [self._DUMMY_CHECKER_CLS]
             await checker.check()
             updates = checker.update_manifests()
 
-            with open(manifest, "r") as f:
+            with open(manifest_path, "r") as f:
                 new_contents = f.read()
 
             self.assertEqual(new_contents, expected_new_contents)
@@ -156,6 +143,34 @@ class TestExternalDataChecker(unittest.IsolatedAsyncioTestCase):
                 self.assertEqual(releases[0].getAttribute("date"), "2019-08-28")
             else:
                 self.assertEqual(releases, [])
+
+
+class TestExternalDataChecker(_TestWithInlineManifest):
+    _DUMMY_CHECKER_CLS = UpdateEverythingChecker
+
+    def setUp(self):
+        init_logging()
+
+    async def test_check_filtered(self):
+        # Use only the URLChecker which is fast so we don't have to wait a lot
+        # for this test; so save the real checkers for later
+        dummy_checker = manifest.ManifestChecker(TEST_MANIFEST)
+        dummy_checker._checkers = [DummyChecker]
+
+        ext_data = await dummy_checker.check()
+        ext_data_from_getter = dummy_checker.get_external_data()
+        self.assertEqual(len(ext_data), len(ext_data_from_getter))
+        self.assertEqual(ext_data, ext_data_from_getter)
+
+        self.assertEqual(len(ext_data), NUM_ALL_EXT_DATA)
+        ext_data = await dummy_checker.check(ExternalData.Type.EXTRA_DATA)
+        self.assertEqual(len(ext_data), NUM_EXTRA_DATA_IN_MANIFEST)
+
+        ext_data = await dummy_checker.check(ExternalData.Type.FILE)
+        self.assertEqual(len(ext_data), NUM_FILE_IN_MANIFEST)
+
+        ext_data = await dummy_checker.check(ExternalData.Type.ARCHIVE)
+        self.assertEqual(len(ext_data), NUM_ARCHIVE_IN_MANIFEST)
 
     async def test_update_json(self):
         filename = "com.example.App.json"
@@ -408,7 +423,7 @@ size: {UpdateEverythingChecker.SIZE}
         )
 
     async def test_check(self):
-        checker = ManifestChecker(TEST_MANIFEST)
+        checker = manifest.ManifestChecker(TEST_MANIFEST)
         ext_data = await checker.check()
 
         self.assertEqual(len(ext_data), NUM_ALL_EXT_DATA)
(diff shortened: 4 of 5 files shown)


tests/test_checker.py
```diff
@@ -135,7 +136,10 @@ async def _test_update(
                 len(checker.get_external_data()),
                 expected_data_count,
             )
-            checker._checkers = [UpdateEverythingChecker]
+
+            if not critical_source_test:
+                checker._checkers = [UpdateEverythingChecker]
```
vchernin: There might be a cleaner way of handling this. The problem is if this line runs with the new tests, some failed assertions will happen. The new tests are still valid though as-is.

 I assume it's because some code here is missing handling for `type: git` sources?

src/lib/externaldata.py
```diff
@@ -45,6 +45,7 @@
     "properties": {
         "type": {"type": "string"},
         "is-main-source": {"type": "boolean"},
+        "is-critical-source": {"type": "boolean"},
```
gasinvein: I'm not a fan of using the term "critical", IMO it's too strong. Maybe "important" would fit better?

src/manifest.py
```diff
@@ -406,6 +407,29 @@ def _update_manifest(self, path, datas, changes):
             log.info("Updating %s", path)
             self._dump_manifest(path)
 
+    def has_version_changed(self, selected_data) -> bool:
```
gasinvein: This would be better be a method (or even a property) on `ExternalBase` objects.

src/manifest.py
```diff
@@ -476,15 +480,42 @@ def update_manifests(self) -> t.List[str]:
...
-            except AppdataNotFound as err:
-                log.info("Not updating appdata: %s", err)
-            except AppdataError as err:
-                self._errors.append(err)
-                log.error(err)
+
+        # if is-critical-source is found, only update the manifest if at least one source with is-critical-source received an update.
+
+        # assume we need to update everything by default
+        critical_version_changed = True
```
gasinvein: Setting `critical_version_changed` to `True` when there are no critical sources is a bit misleading. Maybe split it to two variables, e.g. `has_critical_source` and `critical_version_changed`? Or make it tri-state, with `None` value indication no critical sources.

src/manifest.py
```diff
@@ -476,15 +480,42 @@ def update_manifests(self) -> t.List[str]:
...
+
+        selected_data = None
```
gasinvein: `selected_data` isn't used outside the loop, so it's not needed.

gasinvein (changes requested):
Thanks implementing this.

On the approach in general: I believe this feature should be gated by some cli or `flathub.json` option, otherwise there is no way to update unimportant sources manually. And if gating it, probably the main source should be considered "critical" by default?

vchernin:
Thanks for the review.

> On the approach in general: I believe this feature should be gated by some cli or `flathub.json` option, otherwise there is no way to update unimportant sources manually. And if gating it, probably the main source should be considered "critical" by default?

What you propose (config option to gate it + re-using main source) is something I considered, but I picked the current approach for a few reasons:

1. With the current approach, if you add `is-critical-source` to a particular source, it is explicit that you have changed something specific to that particular source. 

This is ideal since someone less familiar with f-e-d-c could (hopefully) understand there is special behaviour by just viewing the manifest, and seeing that some sources are marked "critical", and therefore are probably being prioritized/treated differently somehow. 

2. The suggestion of a config option in `flathub.json` + the last source guessed as the main/critical one isn't very intuitive IMO. Mainly since one would have to read the docs to learn that the last source is the main/critical one. 

3. Also with the suggestion, it is not obvious how one would add a second critical source if wanted (since the first critical source is guessed/reused from `is-main-source`). While with the current approach, if the first source is marked critical one would probably guess the second one should be given `is-critical-source` as well.

4. Finally with the current approach, requiring `is-critical-source` present in the manifest is itself a gate for this feature. So no need to enable it in `flathub.json` or elsewhere (which I think is best).

I do think a CLI option to update unimportant sources manually would be useful, perhaps `--always-update-manifest`?

src/lib/externaldata.py
vchernin: I have no real opinion on this, indeed important could be used instead.

I wonder though, is there a better parameter name than `is-important-source`? Is it possible to somehow communicate the changed manifest editing/PR behaviour as part of the option? (Since at its core that's what this behaviour change is about).

e.g. `require-update-for-manifest-edits` or something? (But that one's a bit long).

gasinvein: I think `is-important` would do.

gasinvein:
> This is ideal since someone less familiar with f-e-d-c could (hopefully) understand there is special behaviour by just viewing the manifest

Explicit is always better than implicit. If there is a gating option like `require-important-update`, there could be no doubt how f-e-d-c will behave. With current approach, there is no indication that adding `is-critical-source`/`is-important` property to any source will change how f-e-d-c updates sources.

My suggestion is as follows:
- if `require-important-update` is `true`, only update the manifest when a source with `is-important` or `is-main-source` is updated
- if `require-important-update` is `false`, always update the manifest
- if `require-important-update` is unset
  - if there is at least one `is-important` source, only update the manifest when a source with `is-important` or `is-main-source` is updated
  - if there are no `is-important` sources, always update the manifest

vchernin:
I like the gating option suggested. It is more complex, but I agree being more explicit with a gating option with the manifest option works well, to avoid unintentionally using this feature.

> update only sources with is-important or is-main-source

Just to be extra clear, the behaviour of this PR was never to update only the "important" sources, but to only save the manifest changes if at least one important source was updated. 

It doesn’t change the checker code, just adds an additional check determining whether or not the manifest should be saved.

That is, if an important source had an update, and also 1 non-important source had an update, f-e-d-c would update both sources together when making PRs/editing the manifest locally.

So the gating option of `update-important-only` would be more accurate as `update-manifest-only-when-an-important-source-updates`. Perhaps the gating option should be renamed to something more explicit (but not as long)?

Am I correct here that you meant the above behaviour I described (and is what I implemented except for the gating option), or did you specifically mean "update only sources with `is-important` or `is-main-source`"?

> > if update-important-only is unset

> if there is at least one is-important source, update only important sources (including main)

I’m not sure I understand the purpose of this scenario. 

My understanding was that the purpose of the gating option was to prevent users changing behaviour if, they added `is-important` without understanding what it meant. But if we do this, since the default is that `update-important-only` is unset, someone could easily accidentally change behaviour by say, copy pasting a module include `is-important` and pasting it in a manifest.

So what I propose is:

- if `update-important-only` is true, update the manifest only when at least one source with `is-important` or `is-main-source` has an update available. In the case no source is marked main or important, retain previous behaviour

- if `update-important-only` is false or unset, update the manifest always (retain previous behaviour)

gasinvein:
> Am I correct here that you meant the above behaviour I described (and is what I implemented except for the gating option), or did you specifically mean "update only sources with `is-important` or `is-main-source`"?

You are correct, sorry for the confusion. I did mean "update the manifest only when a `is-important` or `is-main-source` source is updated". Edited my post accordingly.

vchernin:
> You are correct, sorry for the confusion. I did mean "update the manifest only when a `is-important` or `is-main-source` source is updated". Edited my post accordingly.

No problem, thanks for confirming. 

With your current post, I still have the same concern about the behaviour if `require-important-update` is unset.

From my previous comment: 
> Since the default is that `require-important-update` is unset, someone could easily accidentally change behaviour by say, copying a module including `is-important` and pasting it in a manifest

Does it not make sense to keep previous behaviour if this feature was not explicitly enabled?

gasinvein:
> With your current post, I still have the same concern about the behaviour if `require-important-update` is unset.
> 
> From my previous comment:
> 
> > Since the default is that `require-important-update` is unset, someone could easily accidentally change behaviour by say, copying a module including `is-important` and pasting it in a manifest
> 
> Does it not make sense to keep previous behaviour if this feature was not explicitly enabled?

Well, you have a point. Let's stick to assuming that `require-important-update` is `false` by default (if unset).

vchernin:
Ok we can do that then. 

Also, do you think an additional CLI force disable option is still beneficial, even though this can be disabled in `flathub.json` ? Something like `--ignore-require-important-update`? While long, that follows the naming pattern.

I would use a CLI option, I don’t like the idea of editing `flathub.json` every time I want to update everything locally (and then remembering to change it back/not commit it). Also, a CLI option is easier than editing `flathub.json` if someone scripts their local setup or something.

So overall I suggest:

- if `require-important-update` is true, only update the manifest when a source with `is-important` or `is-main-source` is updated. If `is-important` and `is-main-source` are not present in the manifest, always update the manifest.
- if `require-important-update` is false or unset, always update the manifest.
- if `--ignore-require-important-update` passed, always update the manifest and ignore `require-important-update`.

gasinvein:
We can add only a CLI option, then read `flathub.json` and pass the CLI option to f-e-d-c in the GitHub workflow.

vchernin:
> We can add only a CLI option, then read flathub.json and pass the CLI option to f-e-d-c in the GitHub workflow.

What you describe is just how to implement the summary in my last comment, or is this an actual user-facing behaviour change? I don’t fully understand what you mean.

gasinvein:
> > We can add only a CLI option, then read flathub.json and pass the CLI option to f-e-d-c in the GitHub workflow.
> 
> What you describe is just how to implement the summary in my last comment, or is this an actual user-facing behaviour change?

No, it's just about how to proceed with implementation.

> I don’t fully understand what you mean.

I mean we can have an option in `flathub.json`, but only read it in the flathub org github workflow, not in f-e-d-c itself.

vchernin:
> We can add only a CLI option, then read flathub.json and pass the CLI option to f-e-d-c in the GitHub workflow.

> I mean we can have an option in flathub.json, but only read it in the flathub org github workflow, not in f-e-d-c itself.

Ok, I think I get it.

What you're saying is f-e-d-c itself will only understand a CLI option. In the case that `flathub.json` exists/is read (e.g. in Flathub's workflow) and if `"require-important-update": "true"` exists in `flathub.json` we need to pass the CLI option `--require-important-update`.

So the CLI options we need:
 - `--require-important-update` (passed by Flathub workflow if  `"require-important-update": "true"` is in `flathub.json`) 

- ~`--ignore-require-important-update` (forces default assumed behaviour if editing locally).~ edit: actually I don't know if there is a use case for this, given the above option.

gasinvein:
> In the case that `flathub.json` exists/is read (e.g. in Flathub's workflow) and if `"require-important-update": "true"` exists in `flathub.json` we need to pass the CLI option `--require-important-update`.

Yep, that was the idea.

vchernin:
Specifically, you mean to add an additional check in `flathub.json` here:

https://github.com/flathub/actions/blob/379a6ec4a04446d6f15c205c6c0c401bc97e8836/flathub-external-data-checker/entrypoint.sh#L12

As opposed to reading in f-e-d-c like this (like for `automerge-flathubbot-prs`):

https://github.com/flathub/flatpak-external-data-checker/blob/89c88675083675ff652a30d95e759a5c9cac1d3a/src/main.py#L254-L259

> Yep, that was the idea.

It makes sense, but what about cases like https://github.com/flathub/com.github.wwmm.easyeffects/blob/master/.github/workflows/update-master.yml? There the flathub-wide action is no longer used, so one would need the CLI option. It works, but I think a json config file is preferable.

Has it ever been considered to make a generic `flatpak.json` to keep generic f-e-d-c options in, and leave the flathub-specific options explictly in `flathub.json`? I think also `automerge-flathubbot-prs` is strangely named and placed in `flathub.json`, when that option works for non flathubbot PRs too.

gasinvein:
> It makes sense, but what about cases like https://github.com/flathub/com.github.wwmm.easyeffects/blob/master/.github/workflows/update-master.yml? There the flathub-wide action is no longer used, so one would need the CLI option. It works, but I think a json config file is preferable.

If one uses a custom workflow to run f-e-d-c, it seems reasonable to pass needed options in the workflow file, no?

> Has it ever been considered to make a generic `flatpak.json` to keep generic f-e-d-c options in, and leave the flathub-specific options explictly in `flathub.json`? I think also `automerge-flathubbot-prs` is strangely named and placed, when that option works for non flathubbot PRs too.

Indeed it feels out of place. This is why I'm considering stopping reading `flathub.json` in f-e-d-c altogether and moving all the `flathub.json` options logic to the gh workflow, but didn't discuss it with flathub folks yet.

vchernin:
> If one uses a custom workflow to run f-e-d-c, it seems reasonable to pass needed options in the workflow file, no?

Yeah, but I think this could be better handled without having to manually pass options. 

Say if making a custom workflow if you're an individual repo. You need to know the "right" CLI options to use (and where to put them). Or, if running an alternative to Flathub you would need to know to reuse the logic in Flathub's workflow to read `$remotename.json`, since you want individual repos to be able to configure options themselves.

A new generic f-e-d-c workflow template (of some kind) replicating some of Flathub's features e.g. includes reading for `$remotename.json`, and understanding `automerge-flathubbot-prs` and `require-important-update` (i.e. passing the options correctly) could help here. 

Then instead of having to copy paste reading `flathub.json` code in every workflow, that would already be handled by the new workflow. So, one wouldn't have to add options to the workflow and could use it mostly as is, since it could read `$remotename.json`.

Then, f-e-d-c could remove the `flathub.json` reading, and instead rely on those running workflows to use the provided `$remotename.json` reading configurations, which would mean everyone should be running f-e-d-c in the same way by default (which I think is cleaner overall).

> Indeed it feels out of place. This is why I'm considering stopping reading flathub.json in f-e-d-c altogether and moving all the flathub.json options logic to the gh workflow, but didn't discuss it with flathub folks yet.

Agreed it should be moved, the above new workflow suggestion I think could potentially help solve that. In any case that sounds like a new issue to open.

vchernin:
All the feeback and suggestions should be implemented now.

I left the commits alone so the changes are still somewhat visible, I can rebase them when the time is right.

I haven't tested with an actual workflow yet. I'm curious do you think `test_checker.py` is reasonable as is, or should the new tests be split off into a new file?

src/lib/externaldata.py
```diff
@@ -185,6 +186,29 @@ def set_new_version(self, new_version: ExternalState, is_update=None):
                 self.state = self.State.BROKEN
             self.new_version = new_version
 
+    def has_version_changed(self) -> bool:
```
gasinvein, line 280: This can be a property.

src/lib/externaldata.py
```diff
@@ -185,6 +186,29 @@ def set_new_version(self, new_version: ExternalState, is_update=None):
...
+        """Detect if some external data had a version change"""
+        update = self.new_version
+        return (
+            update is not None
+            and update.version is not None
+            and (
+                (
+                    isinstance(update, ExternalFile)
```
gasinvein: Ideally, instead of checking type of something on `self`, `has_version_changed` method can be different on `ExternalData` and `ExternalGitRepo` classes.

gasinvein (commented):
> I'm curious do you think `test_checker.py` is reasonable as is, or should the new tests be split off into a new file?

Probably so, at least split it off to another test class.

Ideally, this test shouldn't hit network, i.e. use some dummy checker that always sets some pre-defined "new version".

vchernin:
Fixed the last two comments.

Moved the tests over to a new file, but as suggested I'd also like to get it working without network, and then I'll cleanup the unecessary parts copied over.

src/lib/checkers.py
```diff
@@ -35,7 +36,11 @@
 class Checker:
     CHECKER_DATA_TYPE: t.Optional[str] = None
     CHECKER_DATA_SCHEMA: t.Dict[str, t.Any]
-    SUPPORTED_DATA_CLASSES: t.List[t.Type[ExternalBase]] = [ExternalData]
+    SUPPORTED_DATA_CLASSES: t.List[t.Type[ExternalBase]] = [
+        ExternalData,
+        ExternalGitRepo,
```
vchernin: Not sure if this is correct, it was necessary to pass this assertion:

https://github.com/flathub/flatpak-external-data-checker/blob/4de54724f00e001df0986a3d01e5575abff1384a/src/lib/checkers.py#L56-L58

I assume this is since my tests happened to be git sources. If it's preferable I can simply move them to html sources or something, it shouldn't matter match.

Though I was curious, why shouldn't this assertion allow a git source? As far as I can tell, the way my git source tests work using this code makes sense (i.e. this code is operating sanely with git sources despite the original code disallowing them).

vchernin:
Tests now run offline as requested.

Took the liberty of rebasing, previous minor fixes didn't seem worth preserving at this point, I hope that's alright.

src/lib/checkers.py
gasinvein: It is incorrect. This `SUPPORTED_DATA_CLASSES` is the default, you shouldn't change it. Instead, you're supposed to redefine it on child classes, like this
```python
class MyCustomChecker(Checker):
    SUPPORTED_DATA_CLASSES = [ExternalData, ExternalGitRepo]

Comment (vchernin) [reply src/lib/checkers.py]: Makes sense, fixed

File: tests/test_checker.py (hunk 1 above)
Comment (vchernin) [reply tests/test_checker.py]: Fixed by moving to a new file

**Comment by vchernin** [comment]
Tested with some PRs, behaves as expected.

File: src/manifest.py [hunk 9]
```diff
@@ -471,20 +459,48 @@ def _update_appdata(self):
...
+
+        found_important_update = None
+
+        if self.require_important_update_enabled:
+
+            for data in self.get_external_data():
+                if data.checker_data.get(IMPORTANT_SRC_PROP) or data.checker_data.get(
+                    MAIN_SRC_PROP
+                ):
+                    selected_data = data
```
gasinvein: Assigning `selected_data` variable is redundant, you can just use `data` inside the loop.

tests/test_importantsource.py
```diff
@@ -0,0 +1,606 @@
...
+        return True
+
+    async def check(self, external_data):
+        logging.debug(
+            "Phony checker checking external data %s and all is always good",
+            external_data.filename,
+        )
+
+
+class UpdateEverythingChecker(DummyChecker):
```
gasinvein: Ideally, the tests for important source should be merged into `test_checker.py` and reusing the common code instead of just copying it over.

vchernin: Hm, I had previously split off the tests into a new file due to https://github.com/flathub/flatpak-external-data-checker/pull/276#pullrequestreview-893415396. 

Would you still prefer re-combining the files (and I assume also the classes)?

The tests combined are about a 1000 lines long. Also, the `DummyChecker` in `test_importantsource.py` subclasses `GitChecker` since the tests I wrote happened to use git sources. 

The tests could be switched to `extra-data` sources. Or, `test_checker.py`'s `DummyChecker` could be switched to subclass `GitChecker` (but I'm unsure if it's 100% correct for the existing `extra-data` tests in `test_checker.py`).

src/manifest.py
vchernin: thanks, fixed

tests/test_importantsource.py
gasinvein: >Would you still prefer re-combining the files (and I assume also the classes)?

My initial suggestion was to move the important source tests into separate class, not entire separate file. I still think it would be better, but no strong opinion on that.

vchernin: > I still think it would be better, but no strong opinion on that.

I've combined the tests files, and split off `_test_update` so it can be reused by both sets of tests. There is now only a few lines of copy paste (the git checker subclassing). I needed to slightly adjust the exisiting `_test_update` calls, but they look correct.

tests/test_checker.py
```diff
@@ -52,6 +54,66 @@
 NUM_NEW_VERSIONS = 2
 
 
+async def _test_update(
```
gasinvein: This is clearly a method, not a standalone function. No point to separate it from the test class. To make it reusable, attach it to a common ancestor of `TestExternalDataChecker` and `TestImportantGitExternalDataChecker`, somehow like this:
```python
class _TestWithInlineManifest(unittest.IsolatedAsyncioTestCase):
    async def _test_update(self, ...):
        ...

class TestExternalDataChecker(_TestWithInlineManifest):
    ...

class TestImportantGitExternalDataChecker(_TestWithInlineManifest):
    ...
```
Then you can call `self._test_update()` on both.

vchernin: thanks, fixed

tests/test_checker.py
```diff
@@ -52,6 +54,67 @@
...
+
+            await checker.check()
+            updates = checker.update_manifests()
+
+            with open(manifest, "r") as f:
+                new_contents = f.read()
+
+            self.assertEqual(new_contents, expected_new_contents)
+            if not test_important_update:
+                self.assertEqual(updates, expected_updates)
```
gasinvein: Making this conditional seems like a bad idea. We should verify that exactly the updates we expected were applied.

tests/test_checker.py
```diff
@@ -52,6 +54,67 @@
...
+            checker = ManifestChecker(manifest, unsafe_mode, require_important_update)
+            self.assertEqual(
+                len(checker.get_external_data()),
+                expected_data_count,
+            )
+
+            if test_important_update:
+                checker._checkers = [GitUpdateEverythingChecker]
+            else:
+                checker._checkers = [UpdateEverythingChecker]
```
gasinvein: Instead of selecting a checker class based on the method argument, you can just declare the checker on the test class, so you can pass it to `checker._checkers` like this:
```suggestion
            checker._checkers = [self._DUMMY_CHECKER_CLS]
```

tests/test_checker.py
```diff
@@ -52,6 +54,67 @@
...
+
+            releases = appdata_doc.getElementsByTagName("release")
+            if new_release:
+                self.assertNotEqual(releases, [])
+                self.assertEqual(releases, releases[:1])
+                self.assertEqual(releases[0].getAttribute("version"), "1.2.3.4")
+                self.assertEqual(releases[0].getAttribute("date"), "2019-08-28")
+            elif (
+                not test_important_update
+            ):  # releases assertions are not needed when testing important updates
```
gasinvein: Why? Checking which source got into appstream release entry seems important.

tests/test_checker.py
```diff
@@ -198,10 +213,10 @@ async def test_update_json(self):
...
 
         await self._test_update(
-            filename,
-            contents,
-            expected_new_contents,
-            ["Update UnityHubSetup.AppImage to 1.2.3.4"],
+            filename=filename,
+            contents=contents,
+            expected_new_contents=expected_new_contents,
+            expected_updates=["Update UnityHubSetup.AppImage to 1.2.3.4"],
```
gasinvein: Please either split unrelated changes to a separate commit or don't include them at all into this PR.
The same applies to other args->kwargs changes below.

gasinvein (changes requested):
Another round of the tests review.

Probably I can address these myself and force-push to your branch, if you don't mind.

vchernin:
> Probably I can address these myself and force-push to your branch, if you don't mind.

Yes, please feel free to force push.

src/main.py
```diff
@@ -388,6 +388,13 @@ def parse_cli_args(cli_args=None):
         help="Enable unsafe features; use only with manifests from trusted sources",
         action="store_true",
     )
+    parser.add_argument(
+        "--require-important-update",
+        help="Require an update to at least one source with is-important or is-main-source to save changes to the manifest."
+        "If no instances of is-important or is-main-source are found, assume normal behaviour and always save changes to the manifest."
```
```suggestion
        help="Require an update to at least one source with is-important or is-main-source to save changes to the manifest. "
        "If no instances of is-important or is-main-source are found, assume normal behaviour and always save changes to the manifest. "
```

src/manifest.py
```diff
@@ -471,20 +459,47 @@ def _update_appdata(self):
...
+
+                    if found_important_update:
+                        log.info(
+                            "Update found for important source: %s, updating manifest",
+                            data,
+                        )
+                        break
+
+        if found_important_update or found_important_update is None:
+
```
wjt:
```suggestion
```

wjt (commented):
Great concept! and I think this design works and avoids having too much noise in the manifest.

vchernin:
the corresponding PR in flathub actions: https://github.com/flathub/actions/pull/22

vchernin:
Tweaked the readme slightly to mention passing the option manually is needed if using a custom workflow. Also, I assume the test commit should be squashed into the main commit?

BTW, maybe some bikeshedding, but I am questioning reading `is-main-source` here. Yes, this feature is opt-in, however reading `is-main-source` could still be slightly confusing. The connection between `require-important-update` and `is-important` is fairly obvious from the name, however the fact that `is-main-source` is also "important" is less clear. Doesn't it seem sane to keep one property corresponding to one behavior change?

Furthermore, best case treating `is-main-source` as "important" would only save one instance of `is-important` from a manifest.

gasinvein:
> The connection between `require-important-update` and `is-important` is fairly obvious from the name, however the fact that `is-main-source` is also "important" is less clear. Doesn't it seem sane to keep one property corresponding to one behavior change?
> 
> Furthermore, best case treating `is-main-source` as "important" would only save one instance of `is-important` from a manifest.

While I agree that `is-main-source: true` implying `is-important: true` isn't very consistent, I believe in 99% cases the main source is also considered important. In the rest 1% cases one can set `is-main-source: true` together with `is-important: false` (seems like a valid test case, btw).

vchernin:
> (seems like a valid test case, btw).

Yup, added a new test case with `is-main-source: true` and `is-important: false`, the manifest logic now requires `is-important` to explictly not be false for a source with `is-main-source: true` to be considered important.

gasinvein (changes requested):
CLI options passed to `ManifestChecker` are now wrapped in a single `CheckerOptions` object, to avoid inflating the `__init__` method. Please rebase and update options propagation accordingly.

vchernin:
> CLI options passed to `ManifestChecker` are now wrapped in a single `CheckerOptions` object, to avoid inflating the `__init__` method. Please rebase and update options propagation accordingly.

Done, also tested on a manifest and everything still seems to work

gasinvein (approved):
I still have a bunch of nitpicks, but it feels like this PR was pending long enough already, so let's merge it.

PR #288, opened 2022-04-07
checkers: Skip download if URL unchanged

### LINKED ISSUE #257: Re-downloading static URLs is unnecessary
We always download upstream file, even if it's not new, i.e. URL was unchanged. This is likely unnecessary in most cases and just wastes network bandwidth. Cases when supposedly the same URL gives changed data are rare and indicate something broken upstream, if this happens - the user may want to manually investigate the problem instead of receiving a PR with checksum update.
So, we should probably skip upstream file download when the URL stayed the same.

gasinvein:
If the new version we've got has the same URL as the current one, don't download it again. This will drastically reduce check times.

Although this means we won't re-validate `extra-data` sources that use checkers that don't always get integrity (e.g. HTMLChecker). There are two ways around this:
- either make an exception for `extra-data` sources in `Checker._update_version()` and still download them
- or allow applying checkers in sequence, so that e.g. URLChecker can still validate `extra-data` after HTMLChecker didn't get a new version

The former is much easier, the later is overall cleaner and more universal. I'd like @wjt opinion on which approach would be better.
Marking as draft until either approach is taken and implemented.

Closes #257

### DIFF
--- a/src/lib/checkers.py
+++ b/src/lib/checkers.py
@@ -151,7 +151,10 @@ class Checker:
         assert latest_version is not None
         assert latest_url is not None
 
-        if latest_url == external_data.current_version.url:
+        if (
+            latest_url == external_data.current_version.url
+            and external_data.type != external_data.Type.EXTRA_DATA
+        ):
             external_data.state |= external_data.State.LATEST
             return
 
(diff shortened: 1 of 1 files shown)


wjt:
I don't think the special-case for extra-data would be too bad. Such sources are meaningfully different because when they are broken the existing version of the app can't be installed, compared to when normal sources are broken.

gasinvein:
Ok, lets just keep re-downloading extra-data sources for now.

PR #290, opened 2022-04-10
Add example workflow to README

vchernin:


README.md
```diff
@@ -69,6 +69,55 @@ options for you and run the image using `podman`:
...
+name: Check for updates
+on:
+  schedule: # for scheduling to work this file must be in the default branch
+  - cron: "0 * * * *" # run every hour
+  workflow_dispatch: # can be manually dispatched under GitHub's "Actions" tab 
+
+jobs:
+  flatpak-external-data-checker:
+    runs-on: ubuntu-latest
+
```
vchernin, line 96: originally this had:
```
container:
      image: ghcr.io/flathub/flatpak-external-data-checker
```
after `runs-on`.

But as far as I can tell, it isn't needed.

README.md
```diff
@@ -69,6 +69,55 @@ options for you and run the image using `podman`:
...
+      matrix:
+        branch: [ master ] # list all branches to check
+    
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ matrix.branch }}
+
+      - uses: docker://ghcr.io/flathub/flatpak-external-data-checker:latest
+        env:
```
vchernin, line 107: does this actually need so many env variables for git?

README.md
```diff
@@ -69,6 +69,55 @@ options for you and run the image using `podman`:
...
+
+To stop Flathub's tool from checking your repo, add `"disable-external-data-checker": true` to `flathub.json` in the default branch.
+
+### Custom workflow
+
+Alternatively, you can use own workflow. This can be useful if e.g. wanting to update non-default branches.
+
+Put this yaml file under `.github/workflows`, e.g. put it in `.github/workflows/update.yaml`. Ensure to put the correct path to the manifest in the last line.
+
+<details><summary>GitHub update workflow</summary><br/>
```
gasinvein: I don't think we need to collapse the snippets. The readme is already long enough and requires ToC to navigate, so the snippets won't change much.

README.md
gasinvein: Indeed, it's not needed unless you want to run `flatpak-external-data-checker` command directly inside the container.

README.md
vchernin: Yeah, I agree it doesn't help much with readme length. Removed the collapsing.

README.md
```diff
@@ -69,6 +69,52 @@ options for you and run the image using `podman`:
...
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ matrix.branch }}
+
+      - uses: docker://ghcr.io/flathub/flatpak-external-data-checker:latest
+        env:
+          GIT_AUTHOR_NAME: Flatpak External Data Checker
+          GIT_COMMITTER_NAME: Flatpak External Data Checker
+          GIT_AUTHOR_EMAIL: 41898282+github-actions[bot]@users.noreply.github.com
```
wjt, line 111: Where does this magic number come from?

vchernin: That email is just github actions' user, the number is the user id. So the [commit author is github actions](https://github.com/vchernin/com.github.wwmm.easyeffects/pull/17/commits/f628ac0e8105f5b101830bf39c5754a786ce6731), not [the user who commited the action file](https://github.com/flathub/com.github.wwmm.easyeffects/pull/127/commits/36d38913628796f85c2e14905ae66c33cda1a1a5).

I found it "documented" [here](https://github.community/t/github-actions-bot-email-address/17204/6). I guess a dummy "flatpakbot" account  might look fancier.

wjt: Indeed it might.

Could you add a comment explaining this?

vchernin: > Could you add a comment explaining this?

done

wjt (approved):

PR #292, opened 2022-04-18
gitchecker: add support for versions property

### LINKED ISSUE #291: Git checker doesn't understand version contraining
Even when restricting the checker to one version, f-e-d-c still suggests the actual latest tag.

```yaml
id: com.virustotal.Uploader
modules:
  - name: extra-cmake-modules
    sources:
      - type: git
        url: https://invent.kde.org/frameworks/extra-cmake-modules.git
        tag: v5.77.0
        commit: 760a013c3ff553b23d456c7787ed76981ec5142f
        x-checker-data:
          type: git
          tag-pattern: ^v(\d[\d.]+\d)$
          versions:
            ==: "1.10.1"
```

```
$ flatpak run org.flathub.flatpak-external-data-checker com.virustotal.Uploader.yaml
INFO    src.manifest: Checking 1 external data items
INFO    src.manifest: Started check [1/1] git extra-cmake-modules.git (from com.virustotal.Uploader.yaml)
INFO    src.lib.externaldata: Source extra-cmake-modules.git: got new version 5.93.0
INFO    src.manifest: Finished check [1/1] git extra-cmake-modules.git (from com.virustotal.Uploader.yaml)
OUTDATED: extra-cmake-modules.git
 Has a new version:
  URL:       https://invent.kde.org/frameworks/extra-cmake-modules.git
  Commit:    eccf695824fa931bca471bad3f1aa3a1f244fdd6
  Tag:       v5.93.0
  Branch:    None
  Version:   5.93.0
  Timestamp: None

INFO    src.main: Check finished with 0 error(s)
```

vchernin:
Fixes https://github.com/flathub/flatpak-external-data-checker/issues/291

### DIFF
--- a/src/checkers/gitchecker.py
+++ b/src/checkers/gitchecker.py
@@ -5,8 +5,9 @@ from distutils.version import LooseVersion
 
 import semver
 
+from ..lib import OPERATORS_SCHEMA
 from ..lib.externaldata import ExternalBase, ExternalGitRepo, ExternalGitRef
-from ..lib.utils import git_ls_remote
+from ..lib.utils import git_ls_remote, filter_versions
 from ..lib.errors import CheckerQueryError, CheckerFetchError
 from ..lib.checkers import Checker
 
@@ -61,6 +62,7 @@ class GitChecker(Checker):
         "type": "object",
         "properties": {
             "tag-pattern": {"type": "string", "format": "regex"},
+            "versions": OPERATORS_SCHEMA,
             "version-scheme": {
                 "type": "string",
                 "enum": list(TAG_VERSION_SCHEMES),
@@ -96,6 +98,8 @@ class GitChecker(Checker):
 
         version_scheme = external_data.checker_data.get("version-scheme", "loose")
         tag_cls = TAG_VERSION_SCHEMES[version_scheme]
+        sort_tags = external_data.checker_data.get("sort-tags", True)
+        constraints = external_data.checker_data.get("versions", {}).items()
 
         matching_tags = []
         refs = await git_ls_remote(external_data.current_version.url)
@@ -114,7 +118,14 @@ class GitChecker(Checker):
             version = tag_match.group(1)
             matching_tags.append(tag_cls(commit, tag, annotated, version))
 
-        if external_data.checker_data.get("sort-tags", True):
+        if constraints:
+            sorted_tags = filter_versions(
+                matching_tags,
+                constraints=constraints,
+                to_string=lambda t: t.version,
+                sort=sort_tags,
+            )
+        elif sort_tags:
             sorted_tags = sorted(matching_tags)
         else:
             sorted_tags = matching_tags
--- a/README.md
+++ b/README.md
@@ -238,6 +238,8 @@ If the project follows [semver](https://semver.org/) specification, you can set
 `version-scheme` property to `semantic` in order to use semantic version scheme for sorting.
 In this case, make sure that `tag-pattern` extracts only valid semver strings.
 
+The [`versions`](#version-constraining) property is supported.
+
 ### JSON checker
 
 The JSON checker allows using [jq](https://stedolan.github.io/jq/) to query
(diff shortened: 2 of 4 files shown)


src/checkers/gitchecker.py
```diff
@@ -119,12 +122,23 @@ async def _check_has_new(external_data: ExternalGitRepo):
...
-                f"{external_data.current_version.url} has no tags matching '{tag_pattern}'"
-            ) from err
+        if constraints:
+            matching_versions = [tag.version for tag in matching_tags]
+            versions = filter_versions(matching_versions, constraints, sort=False)
+            try:
+                latest_version = versions[-1]
+                for found_tag in matching_tags:
+                    if found_tag.version is latest_version:
+                        latest_tag = found_tag
```
gasinvein: `filter_versions()` supports `to_string` arg. If you pass a callable to it, you can filter `TagWithVersion` objects directly, without converting them to version string list first.
E.g. `filter_versions(matching_tags, to_string = lambda t: t.version)` should work.
This way you won't need to iterate over `matching_tags` again, looking for a tag with matching version.

gasinvein: Overall, I'd change it like this:
```patch
diff --git a/src/checkers/gitchecker.py b/src/checkers/gitchecker.py
index f4d5bea..3d5db67 100644
--- a/src/checkers/gitchecker.py
+++ b/src/checkers/gitchecker.py
@@ -98,6 +98,7 @@ class GitChecker(Checker):
 
         version_scheme = external_data.checker_data.get("version-scheme", "loose")
         tag_cls = TAG_VERSION_SCHEMES[version_scheme]
+        sort_tags = external_data.checker_data.get("sort-tags", True)
         constraints = external_data.checker_data.get("versions", {}).items()
 
         matching_tags = []
@@ -117,28 +118,24 @@ class GitChecker(Checker):
             version = tag_match.group(1)
             matching_tags.append(tag_cls(commit, tag, annotated, version))
 
-        if external_data.checker_data.get("sort-tags", True):
+        if constraints:
+            sorted_tags = filter_versions(
+                matching_tags,
+                constraints=constraints,
+                to_string=lambda t: t.version,
+                sort=sort_tags,
+            )
+        elif sort_tags:
             sorted_tags = sorted(matching_tags)
         else:
             sorted_tags = matching_tags
 
-        if constraints:
-            matching_versions = [tag.version for tag in matching_tags]
-            versions = filter_versions(matching_versions, constraints, sort=False)
-            try:
-                latest_version = versions[-1]
-                for found_tag in matching_tags:
-                    if found_tag.version is latest_version:
-                        latest_tag = found_tag
-            except IndexError as err:
-                raise CheckerQueryError("Can't find matching version") from err
-        else:
-            try:
-                latest_tag = sorted_tags[-1]
-            except IndexError as err:
-                raise CheckerQueryError(
-                    f"{external_data.current_version.url} has no tags matching '{tag_pattern}'"
-                ) from err
+        try:
+            latest_tag = sorted_tags[-1]
+        except IndexError as err:
+            raise CheckerQueryError(
+                f"{external_data.current_version.url} has no tags matching '{tag_pattern}'"
+            ) from err
 
         new_version = ExternalGitRef(
             url=external_data.current_version.url,

```

vchernin: Thanks, that is much cleaner.

gasinvein:
Looks good, can you please also add a note about version constraining to the readme, similar to this?
https://github.com/flathub/flatpak-external-data-checker/blob/dfac7bc79493a9380c34a134b1eaf05c6eefece4/README.md?plain=1#L308

vchernin:
> Looks good, can you please also add a note about version constraining to the readme, similar to this?
> 
> https://github.com/flathub/flatpak-external-data-checker/blob/dfac7bc79493a9380c34a134b1eaf05c6eefece4/README.md?plain=1#L308

Done

gasinvein:
Thanks!

PR #297, opened 2022-04-29
checkers: ability to replace timestamp

catsout:
add 'timestamp' parameter to _update_version, for jsonchecker using timestamp from json rather than using the one from url response date.  

Fix for tarballs from GitHub.  

Example:
```yaml
    sources:
      - type: archive
        url: https://github.com/laurent22/joplin/archive/v2.8.4.tar.gz
        sha256: d039577ecc1e7589e7432f45d4e12308d16b5df0c1b0cc928bc451d8136761d4
        x-checker-data:
          type: json
          url: https://api.github.com/repos/laurent22/joplin/releases
          version-query: .[0] | .tag_name | sub("^v"; "")
          url-query: '"https://github.com/laurent22/joplin/archive/v" + $version +
            ".tar.gz"'
          timestamp-query: .[0] | .published_at
```

### DIFF
--- a/tests/io.github.stedolan.jq.yml
+++ b/tests/io.github.stedolan.jq.yml
@@ -10,6 +10,17 @@ modules:
           url: https://api.github.com/repos/stedolan/jq/releases/latest
           version-query: '.tag_name | sub("^jq-"; "")'
           url-query: '.assets[] | select(.name=="jq-" + $version + ".tar.gz") | .browser_download_url'
+
+      - type: archive
+        dest-filename: jq-1.4.tarball.tar.gz
+        url: https://api.github.com/repos/stedolan/jq/tarball/jq-1.4
+        sha256: "0000000000000000000000000000000000000000000000000000000000000000"
+        x-checker-data:
+          type: json
+          url: https://api.github.com/repos/stedolan/jq/releases/13660432
+          version-query: '.tag_name | sub("^jq-"; "")'
+          url-query: '.tarball_url'
+          timestamp-query: '.published_at'
     modules:
 
       - name: oniguruma
--- a/tests/test_jsonchecker.py
+++ b/tests/test_jsonchecker.py
@@ -18,7 +18,7 @@ class TestJSONChecker(unittest.IsolatedAsyncioTestCase):
         checker = ManifestChecker(TEST_MANIFEST)
         ext_data = await checker.check()
 
-        self.assertEqual(len(ext_data), 7)
+        self.assertEqual(len(ext_data), 8)
         for data in ext_data:
             self.assertIsNotNone(data)
             if data.filename == "jq-1.4.tar.gz":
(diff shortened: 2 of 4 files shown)


wjt:
Please could you add a test and update the documentation in README.md?

README.md
```diff
@@ -315,6 +315,7 @@ for git type sources, specify tag query and, optionaly, commit and version queri
 ```
 
 `timestamp-query` is optional, but if provided - must return a string with timestamp in ISO format.
+`timestamp-query` is also available on non-git type sources.
```
gasinvein: I believe this is implied, and the current (pre-this-PR) behaviour is a flaw. So no need to state this explicitly.

gasinvein (changes requested):
Thanks for addressing this.
Please split changes to `checkers` and `jsonchecker` into different commits (changes to tests should go into the same commit that has added changes being tested).

tests/test_jsonchecker.py
```diff
@@ -38,6 +38,11 @@ async def test_check(self):
                         sha256="0000000000000000000000000000000000000000000000000000000000000000"
                     ),
                 )
+            elif data.filename == "jq-1.4.tarball_url.tar.gz":
```
gasinvein: Should match `dest-filename` in the manifest, I suppose:
```suggestion
            elif data.filename == "jq-1.4.tarball.tar.gz":
```

tests/test_jsonchecker.py
```diff
@@ -38,6 +38,11 @@ async def test_check(self):
...
+                self.assertNotEqual(
```
gasinvein: Since we get specific release date from GH API, can we compare to it? I.e. use `assertEqual` instead of `assertNotEqual` here.

tests/test_jsonchecker.py
```diff
@@ -38,6 +38,11 @@ async def test_check(self):
                         sha256="0000000000000000000000000000000000000000000000000000000000000000"
                     ),
                 )
+            elif data.filename == "jq-1.4.tarball.tar.gz":
+                self.assertEqual(
+                    data.new_version.timestamp,
+                    datetime.fromisoformat("2018-11-02T01:54:23+00:00"),
```
gasinvein:
```suggestion
                    datetime.datetime.fromisoformat("2018-11-02T01:54:23+00:00"),
```

tests/io.github.stedolan.jq.yml
```diff
@@ -10,6 +10,17 @@ modules:
...
+      - type: archive
+        dest-filename: jq-1.4.tarball.tar.gz
+        url: https://api.github.com/repos/stedolan/jq/tarball/jq-1.4
+        sha256: "0000000000000000000000000000000000000000000000000000000000000000"
+        x-checker-data:
+          type: json
+          url: https://api.github.com/repos/stedolan/jq/releases/13660432
+          version-query: '.tag_name | sub("^jq-"; "")'
+          url-query: '.tarball_url'
+          timestamp: '.published_at'
```
gasinvein:
```suggestion
          timestamp-query: '.published_at'
```

gasinvein (changes requested):

catsout:
BTW, `./run-in-container.sh python3 -m unittest discover` is not working on my machine.

vchernin:
> BTW, `./run-in-container.sh python3 -m unittest discover` is not work on my machine.

https://github.com/flathub/flatpak-external-data-checker/pull/275 fixes it

gasinvein (approved):
LGTM, thanks.

Issue #144, opened 2021-03-02
Update only when two sources are at the same version number


TheEvilSkeleton:
I'm not sure exactly how to explain this, but the [com.github.Eloston.UngoogledChromium](https://github.com/flathub/com.github.Eloston.UngoogledChromium) flatpak Chromium *and* ungoogled-chromium patches require to be at the same version. See:
- https://github.com/flathub/com.github.Eloston.UngoogledChromium/blob/master/com.github.Eloston.UngoogledChromium.yaml#L176
- https://github.com/flathub/com.github.Eloston.UngoogledChromium/blob/master/com.github.Eloston.UngoogledChromium.yaml#L253

Since Chromium (the latter) comes with updates sooner than ungoogled-chromium's patches, flathubbot starts building the ungoogled-chromium flatpak without the new patches. See https://github.com/flathub/com.github.Eloston.UngoogledChromium/pull/18.

How can I make flathubbot only PR when both versions are the same?

gasinvein:
>How can I make flathubbot only PR when both versions are the same?

You can't, sources are independent from each other. Although it makes sense to implement interconnection between modules. I'm not yet sure how to approach this, so ideas welcome.

PR #302, opened 2022-05-22
Implement parent-child relations between sources

gasinvein:
This adds foundation for resolving #144

Allow the user to specify `parent-id` property in `x-checker-data`, indicating that the current source will use check results from the source with specified id.

In order to establish inter-source relations, we need some kind of source identifiers that are not changed with updates. For that, we can either use a user-provided id, or construct id from the module name and the source position number. For that, we collect flatpak-builder modules.

In scope of this PR, inter-source relations only affect checking order (child check starts only after its parent was checked). But from now on, checkers can take parent source data into account (e.g. access `external_data.parent.new_version`) and, depending on it, pick different upstream url or pre-populate git tag to match parent, for example. Actual parent data usage is to be implemented on per-checker basis.

### DIFF
--- a/src/lib/externaldata.py
+++ b/src/lib/externaldata.py
@@ -37,6 +37,7 @@ from .errors import (
 )
 from .checksums import MultiDigest
 
+_BM = t.TypeVar("_BM", bound="BuilderModule")
 _BS = t.TypeVar("_BS", bound="BuilderSource")
 _ES = t.TypeVar("_ES", bound="ExternalState")
 
@@ -54,6 +55,32 @@ CHECKER_DATA_SCHEMA_COMMON = {
 }
 
 log = logging.getLogger(__name__)
+
+
+@dataclasses.dataclass
+class BuilderModule:
+    name: str
+    module: t.Dict[str, t.Any]
+    module_path: str
+    parent: t.Optional[BuilderModule] = None
+    sources: t.List[BuilderSource] = dataclasses.field(default_factory=lambda: [])
+
+    @classmethod
+    def from_manifest(
+        cls: t.Type[_BM],
+        module_path: str,
+        module: t.Dict,
+        parent: t.Optional[BuilderModule] = None,
+    ) -> _BM:
+        return cls(
+            name=module["name"],
+            module=module,
+            module_path=module_path,
+            parent=parent,
+        )
+
+    def __str__(self):
+        return self.name
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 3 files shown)


gasinvein:
@wjt Got time to review this? I'd like to submit some features that depend on mechanics introduced by this PR, in order get rid of [one](https://github.com/flathub/org.telegram.desktop/blob/5dbb49b5d44e678ed11cddfe562dc864a046a727/.github/workflows/tg_owt-updater.yml) of the custom GH workflows in Telegram flatpak.

wjt:
Sorry, I don't think I will have time this week or next...

gasinvein:
Oh, ok, understood. I doubt this PR alone could introduce regressions, but it's still pretty big. Can you at least tell if you agree with the approach in general, i.e. allowing sources to reference one another and letting checkers use check result data from "parent" source?

wjt:
I wonder if you can reuse the existing structure of manifests, both to avoid needing to define a new ad-hoc system of source IDs, and to remove the possibility of cycles.

One thought is that we could restore something similar to the module-level check that was dropped in #114. Then in Ungoogled Chromium, either move the `ungoogled-chromium` module's single source to be a source of the `chromium` module below, or use the fact that modules can be nested to make it a child. Then the parent-child relationship is established by the nesting of the manifest, there is no possibility of cycles.

A cousin of this approach would be to only allow sources to refer to (a) sources of parent modules; or (b) sibling sources that are earlier in the list.

I wonder if it would be worth giving sources an `"id"` property in flatpak-builder itself. Or using `"x-id"` in the source, rather than the ID being in the checker data. Maybe not.

gasinvein:
> remove the possibility of cycles

FWIW cycles are explicitly checked for. Although I agree that avoiding cycles naturally would've been better, just don't really see a way to.

> use the fact that modules can be nested to make it a child. Then the parent-child relationship is established by the nesting of the manifest, there is no possibility of cycles.

I like this idea, but I guess the nested (lower-level) module should be considered a parent, since the upper-level one depends on it and is built after it. E.g. here
```yaml
name: child
...
modules:
  - name: parent1
    ...
    modules:
      - name: grandparent
        ...
  - name: parent2
    ...
```
it would make more sense for the `child` module source checker to depend on the `parent1` module source check result, not the other way round. Right? Not really sure about that.

How do we determine which one of the parent module sources should be used? Also with f-b modules can have multiple parents.

Another example: in the Wine flatpak, we have `wine-gecko` and `wine-mono` modules, both would depend on the check result from `wine` module source, but both should be built _after_ `wine`. It doesn't seem possible to achieve with f-b module nesting alone.

gasinvein:
I didn't manage to figure out a way to avoid adding ad-hoc source id system, but I agree avoiding it would be nice.
Although I don't expect many apps to use this source relationship feature, so if someone comes up with a better idea of establishing source ids in the future, we can switch to such new mechanism and help existing apps migrate.

PR #305, opened 2022-06-15
JSONChecker: Support YAML URLs and access to parent source properties

gasinvein:
These two features are distinct, but coupled together because they depend on the same prior changes, namely reading JSON objects on Python side instead of passing raw bytes to JQ.


gasinvein:
>`src/lib/externaldata.py:210: error: Unsupported left operand type for | ("Dict[str, Any]")`

Ugh. Mypy passes fine on my machine. Probably Python is too old on the GH runners?

PR #308, opened 2022-06-19
htmlchecker: Allow using parent source vars in url

gasinvein:
This makes HTMLChecker's html `url` a template that can be substituted with parent source properties.
Since there is no way to use `or` statements in a string template to alternate between current and new parent source state, we have to do this at checker side: if there is a new version - use it, otherwise - fall back to current version.


gasinvein:
I'm still not sure about the implemented way to access parent source properties from a child x-checker-data, so I'm leaving this as undocumented feature for now. Let's see how it goes for some apps on flathub that I myself maintain.

Issue #315, opened 2022-07-11
flathubbot is spamming the Buildserver with org.kde.kontact


JakobDev:
I saw org.kde.kontact multiple times on the Buildserver ons serveral days. I wondred why and took a look at the [PRs of org.kde.kontact](https://github.com/flathub/org.kde.kontact/pulls). f-e-d-c currently creates a PR almost every time when it runs, which is hourly, what results in a lot of testbuilds. The Build fails, but the Buildserver did not recognize that, what causes that the Build is hanging. At the time of writing, there are currently 3 testbuilds of org.kde.kontact on the Buildserver, which may block other Apps. You should take a look at what causes this Spam.

gasinvein:
In the log there are errors like this:
>`ERROR   src.manifest: Failed to check archive akregator/akregator-22.04.2.tar.xz with AnityaChecker: Wrong content type 'text/plain' received from 'https://download.kde.org/stable/release-service/22.04.3/src/akregator-22.04.3.tar.xz'`

It indicates a broken or outdated mirror. And updating this many sources at once makes hitting a broken mirror highly likely. This, in turn, causes the set of successfully updated sources to be randomized on each run, resulting in PR spam.

Honestly, I'm not sure what we can do about this. Hardcode some "primary" mirror instead of `download.kde.org`, if KDE provides one?

tinywrkb:
Can we at least disable the flathubbot checks for it until the maintainer will update the app? I mean, it's still generating new PRs.

gasinvein:
It can be disabled by setting `"disable-external-data-checker": true` in `flathub.json`. From the way you put it, I suppose the maintainers are AFK currently?

JakobDev:
If f-e-d-c has a broken mirror in one run, but it worked and found a new version in a previous run, it can use the version from the previous run. That could stop such a Spam in the future.

Tamaranch:
I have a duplicate PR and I guess it's the same problem: https://github.com/flathub/org.xfce.ristretto/pull/81 and https://github.com/flathub/org.xfce.ristretto/pull/82

> ERROR   src.manifest: Failed to check git libopenraw/libopenraw.git with AnityaChecker: Listing Git remote failed: Command '['git', 'ls-remote', '--exit-code', 'https://github.com/freedesktop/libopenraw.git']' returned non-zero exit status 128.

However in this case there is only one updated package and only one dependency whose url is broken, so I'm not sure if the above remark is still relevant:

[...]

PR #316, opened 2022-07-19
Limit simultaneous connections

gasinvein:
Should prevent some of the issue similar to #315.

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -54,7 +54,6 @@ from .checkers import ALL_CHECKERS
 
 MAIN_SRC_PROP = "is-main-source"
 IMPORTANT_SRC_PROP = "is-important"
-MAX_MANIFEST_SIZE = 1024 * 100
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


src/manifest.py
```diff
@@ -75,8 +74,10 @@ def find_appdata_file(directory, appid):
 @dataclasses.dataclass(frozen=True)
 class CheckerOptions:
     allow_unsafe: bool = False
-    max_manifest_size: int = MAX_MANIFEST_SIZE
+    max_manifest_size: int = 1024 * 100
     require_important_update: bool = False
+    conn_limit_per_host: int = 5
+    conn_limit: int = 50
```
gasinvein, line 80: @nanonyme IIRC you were the one who previously suggested to limit simultaneousl connections to prevent upstreams from throttling us. Does this seem like a reasonable default?

nanonyme: @gasinvein in general you're better off having one connection per backend and using HTTP keepalive. TLS handshakes are *slow*.

gasinvein: I'm afraid 1 connection per host will be way too slow: in many apps, there are several sources from GitHub.

PR #317, opened 2022-07-27
Document that tag-pattern of git checker is not needed in some cases

JakobDev:


README.md
```diff
@@ -289,6 +289,9 @@ exactly one match group (the pattern will be used to extract version from tag):
 By default tags are sorted based on version number extracted from tag.
 To disable sorting and keep order from `git ls-remote`, set `sort-tags` to `false`.
 
+If the tags are just the version without anything else e.g. `1.0`, you don't need to
```
gasinvein, line 292: It also matches versions prefixed with `v`, like `v1.0`

PR #320, opened 2022-08-08
README: Document write access for Flatpak version

wjt:
This is a common trap for the unwary. Strictly speaking the `--filesystem` flag
is only needed when running with `--update` or similar but it's easier to just
document it as always needed.

Another nicety not implemented here would be to detect this situation and give a hint in the error message, rather than just dying with "permission denied".

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ flatpak install --from https://dl.flathub.org/repo/appstream/org.flathub.flatpak
 And run with
 
 ```bash
-flatpak run org.flathub.flatpak-external-data-checker MANIFEST_FILE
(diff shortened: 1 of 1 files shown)


gasinvein:
Maybe we should do a run-time check if the manifest file is writable, and suggest user to pass `--filesystem=...` if it isn't?

gasinvein:
Anyway, until that's implemented, let's update the readme.

dreua:
See https://github.com/flathub/org.flathub.flatpak-external-data-checker/issues/117 and #390 for current work on the remaining tasks here regarding a more helpful error message.

And a late thank you for fixing the documentation!

PR #321, opened 2022-08-08
Consistently default stable-only to true

wjt:
Previously, gnomechecker would default stable-only to true, but
anityachecker and pypichecker defaulted it to false.

true is the sensible default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,7 +368,8 @@ and add a template for source download URL.
     "url-template": "https://github.com/flatpak/flatpak/releases/download/$version/flatpak-$version.tar.xz"
 }
 ```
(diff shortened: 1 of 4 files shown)


gasinvein:
Makes sense.

Codepath where `stable_only` is `False` is uncovered now. Can you please set `stable-only: false` on some of the modules in `tests/org.flatpak.Flatpak.yml` where `stable-only` isn't set?

wjt:
How's that?

gasinvein:
Now the codepath where neither constraints are set, nor `stable-only` is requested, is uncovered. But it's dead simple anyway (there we just get a single `latest_version` field from Anitya), should be fine.

PR #327, opened 2022-09-05
gnomechecker: Don't crash if no versions satisfy constraint

### LINKED ISSUE #326: Gnome checker: exits with an error when applied versions filter set to target a future release
Similar to #240, but here with the Gnome checker.  
I'm using the latest [org.flathub.flatpak-external-data-checker](https://github.com/flathub/org.flathub.flatpak-external-data-checker) from Flathub.

Meson buildsystem is broken in the uploaded dist release tarball on Gnome downloads server (reported, dist target fix, [no updated tarball](https://gitlab.freedesktop.org/cairo/cairo/-/issues/548)), so I have to target a future release with the version filter, and in the meantime use the auto-generated tarball.

`cairo.yaml`
```yaml
name: cairo-static
sources:
  - type: archive
    url: https://gitlab.freedesktop.org/cairo/cairo/-/archive/1.17.6/cairo-1.17.6.tar.gz
    sha256: a2227afc15e616657341c42af9830c937c3a6bfa63661074eabef13600e8936f
    x-checker-data:
      type: gnome
      name: cairo
      # NOTE: 1.17.6 tarball is broken on gnome.org
      versions:
        '>': 1.17.6
```

Output
```
$ flatpak-external-data-checker cairo.yaml

INFO    src.manifest: Checking 1 external data items
INFO    src.manifest: Started check [1/1] archive cairo-static/cairo-1.17.6.tar.gz (from cairo.yaml)
  File "/app/lib/fedc/src/checkers/gnomechecker.py", line 75, in check
    latest_version = list(filter(_is_stable, filtered_versions))[-1]
IndexError: list index out of range

During handling of the above exception, another exception occurred:

    main()
  File "/app/lib/fedc/src/main.py", line 457, in main
    return future.result()
  File "/app/lib/fedc/src/main.py", line 420, in run_with_args
  File "/app/lib/fedc/src/manifest.py", line 409, in check
  File "/app/lib/fedc/src/manifest.py", line 344, in _check_data
    await checker.check(data)
  File "/app/lib/fedc/src/checkers/gnomechecker.py", line 77, in check
    latest_version = filtered_versions[-1]
IndexError: list index out of range
```

wjt:
Fixes #326

### DIFF
--- a/src/checkers/gnomechecker.py
+++ b/src/checkers/gnomechecker.py
@@ -70,18 +70,21 @@ class GNOMEChecker(Checker):
         if constraints:
             filtered_versions = filter_versions(filtered_versions, constraints)
 
-        if stable_only:
-            try:
-                latest_version = list(filter(_is_stable, filtered_versions))[-1]
-            except IndexError:
+        try:
+            if stable_only:
+                try:
+                    latest_version = list(filter(_is_stable, filtered_versions))[-1]
+                except IndexError:
+                    latest_version = filtered_versions[-1]
+                    log.warning(
+                        "Couldn't find any stable version for %s, selecting latest %s",
+                        project_name,
+                        latest_version,
+                    )
+            else:
                 latest_version = filtered_versions[-1]
-                log.warning(
-                    "Couldn't find any stable version for %s, selecting latest %s",
-                    project_name,
-                    latest_version,
-                )
-        else:
-            latest_version = filtered_versions[-1]
+        except IndexError as e:
+            raise CheckerQueryError("No matching versions") from e
 
         proj_files = downloads[project_name][latest_version]
 
--- a/tests/org.gnome.baobab.json
+++ b/tests/org.gnome.baobab.json
@@ -64,6 +64,23 @@
                     }
                 }
             ]
+        },
+        {
+            "name" : "cairo-static",
+            "sources" : [
+                {
+                    "type": "archive",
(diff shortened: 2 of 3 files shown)


src/checkers/gnomechecker.py
```diff
@@ -70,18 +70,21 @@ async def check(self, external_data: ExternalBase):
...
-                latest_version = list(filter(_is_stable, filtered_versions))[-1]
-            except IndexError:
+        try:
+            if stable_only:
+                try:
+                    latest_version = list(filter(_is_stable, filtered_versions))[-1]
+                except IndexError:
+                    latest_version = filtered_versions[-1]
+                    log.warning(
+                        "Couldn't find any stable version for %s, selecting latest %s",
```
wjt, line 80: What's the rationale for this (existing) behaviour?

gasinvein: IIRC it had something to do with the `_is_stable` logic being unreliable in general.
I guess we can drop this warning and just error out.

wjt: Dropping it made a test fail, and in the past 2 weeks I've not had time to get back to this! I think I will leave this existing behaviour for now.

gasinvein: Ok, let's leave it as is for now.

gasinvein (approved):

PR #332, opened 2022-10-13
Show base branch in PR title

JakobDev:
f-e-d-c is currently used by [com.riverbankcomputing.PyQt.BaseApp](https://github.com/flathub/com.riverbankcomputing.PyQt.BaseApp) on 3 different branches. Having the branch (e.g. 5.15-22.08 or 6.3) would make it possible to see which branch is targeted directly in the Notification or the PR list.

Note: This is currently not tested

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -438,8 +438,13 @@ def open_pr(
         log_url = f"https://github.com/{gh_repo_name}/actions/runs/{gh_run_id}"
         pr_message += f"\n\n[📋 View External data checker logs]({log_url})"
 
+    subject = change.subject or ""
+
+    if base != origin_repo.default_branch:
+        subject = f"[{base}] {subject}"
+
     pr = origin_repo.create_pull(
-        change.subject,
+        subject,
         pr_message,
         base,
         head,
--- a/src/main.py
+++ b/src/main.py
@@ -444,10 +444,10 @@ def open_pr(
         subject = f"[{base}] {subject}"
 
     pr = origin_repo.create_pull(
-        subject,
-        pr_message,
-        base,
-        head,
+        title=subject,
+        body=pr_message,
+        base=base,
+        head=head,
         maintainer_can_modify=True,
     )
     log.info("Opened pull request %s", pr.html_url)


src/main.py
```diff
@@ -247,6 +248,11 @@ def open_pr(
 
     head = "{}:{}".format(repo.owner.login, change.branch)
     pr_message = ((change.body or "") + "\n\n" + DISCLAIMER).strip()
+    
+    if branch_title:
```
wjt: or something like:

```suggestion
    if base not in ('master', 'beta'):
```

JakobDev: I personal prefer a argument, so the User can decide if he want it or not. There may be someone who uses f-e-d-c on the master and the beta branch and want to know, if a PR goes against the master or the beta branch.

gasinvein:
I wonder if it could be more generic - instead of a boolean option determining whether or not to put branch name into PR title, use a string option with PR title template, with `$updated` as the default, and ability to set it to something like `[$branch] $updated`

src/main.py
```diff
@@ -210,6 +210,7 @@ def open_pr(
     change: CommittedChanges,
     manifest_checker: manifest.ManifestChecker = None,
     fork: t.Optional[bool] = None,
+    branch_title: t.Optional[bool] = None,
```
gasinvein:
```suggestion
    branch_title: bool = False,
```

bbhtt:
Is this still needed? Instead of having branch in the title, labels would be better I think. fedc as of yesterday supports setting PR labels. Also the base branch is included in the branches it creates

JakobDev:
I'm already using labels. The problem with labels is that they don't show up in Notifications.

bbhtt:
Ok so can you rebase and fix the conflict? IMO these minor things don't need be an argument and can be added by default to the title but if you prefer then it's ok.

JakobDev:
> can be added by default to the title

So just add it to the title when the branch is not master?

bbhtt:
You can add it even when it is master. There is no reason to special case a single branch.

PR #335, opened 2022-11-27
Detect content type from magic number of first chunk

### LINKED ISSUE #245: Can't download files with `Content-Type` improperly set to `text/*`
This is largely expected fallout from #219

Currently we're getting "Wrong content type" error from these sources:
```
2021-10-14T12:46:51.4076530Z ERROR   src.checker: Failed to check archive pinentry-1.1.0.tar.bz2 with AnityaChecker: Wrong content type 'text/plain' received from 'https://www.gnupg.org/ftp/gcrypt/pinentry/pinentry-1.2.0.tar.bz2'
2021-10-14T12:47:25.9482623Z ERROR   src.checker: Failed to check archive QScintilla-2.11.6.tar.gz with AnityaChecker: Wrong content type 'text/plain' received from 'https://www.riverbankcomputing.com/static/Downloads/QScintilla/2.13.0/QScintilla_src-2.13.0.tar.gz'
2021-10-14T12:47:59.2745988Z ERROR   src.checker: Failed to check archive gtk-vnc-1.2.0.tar.xz with AnityaChecker: Wrong content type 'text/plain' received from 'https://download.gnome.org/sources/gtk-vnc/1.2/gtk-vnc-1.2.0.tar.xz'
2021-10-14T12:48:55.7106761Z ERROR   src.checker: Failed to check archive libpcap-1.10.1.tar.gz with AnityaChecker: Wrong content type 'text/html; charset=UTF-8' received from 'http://www.tcpdump.org/release/libpcap-1.10.1.tar.gz'
```

Alexander-Wilms:
Should fix #245

For some reason, tests.test_checker.TestExternalDataChecker finds 3 new versions instead of 2, even without my changes applied: See the latest CI job of this branch, which prints the found sources:

https://github.com/Alexander-Wilms/flatpak-external-data-checker/tree/investigate-test-test_checker_testexternaldatachecker

### DIFF
--- a/dependencies.apt.txt
+++ b/dependencies.apt.txt
@@ -21,6 +21,7 @@ python3-jsonschema
 python3-editorconfig
 python3-lxml
 python3-packaging
+python3-magic
 squashfs-tools
 ssh-client
 jq
--- a/src/lib/utils.py
+++ b/src/lib/utils.py
@@ -40,6 +40,7 @@ from ruamel.yaml import YAML
 from elftools.elf.elffile import ELFFile
 import aiohttp
 import editorconfig
+import magic
 
 from . import externaldata, TIMEOUT_CONNECT, HTTP_CHUNK_SIZE, OPERATORS
 from .errors import CheckerRemoteError, CheckerQueryError, CheckerFetchError
@@ -116,19 +117,26 @@ async def get_extra_data_info_from_url(
         real_url = str(response.url)
         info = response.headers
 
-        content_type = info.get(aiohttp.hdrs.CONTENT_TYPE)
-        if (
-            content_type is not None
-            and content_type_deny is not None
-            and any(r.match(content_type) for r in content_type_deny)
-        ):
-            raise CheckerFetchError(
-                f"Wrong content type '{content_type}' received from '{url}'"
+        def content_type_rejected(content_type: t.Optional[str]) -> bool:
+            return (
+                content_type is not None
+                and content_type_deny is not None
+                and any(r.match(content_type) for r in content_type_deny)
             )
 
         checksum = MultiHash()
+        first_chunk = True
         size = 0
         async for chunk in response.content.iter_chunked(HTTP_CHUNK_SIZE):
+            if first_chunk:
+                first_chunk = False
+                # determine content type from magic number since http header may be wrong
+                actual_content_type = magic.from_buffer(chunk, mime=True)
+                if content_type_rejected(actual_content_type):
+                    raise CheckerFetchError(
+                        f"Wrong content type '{actual_content_type}' received from '{url}'"
+                    )
+
             checksum.update(chunk)
             size += len(chunk)
             if dest_io is not None:
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -29,7 +29,7 @@ from tempfile import TemporaryDirectory
 
 import aiohttp
 
-from src.lib.errors import CheckerError
+from src.lib.errors import CheckerFetchError
 from src.lib.utils import (
     parse_github_url,
     strip_query,
@@ -222,12 +222,34 @@ class TestDownload(unittest.IsolatedAsyncioTestCase):
     async def asyncTearDown(self):
         await self.http.close()
 
+    async def test_correct_content_type(self):
+        await get_extra_data_info_from_url(
+            url=f"https://ftp.gnu.org/gnu/gzip/gzip-1.12.tar.gz",
+            session=self.http,
+            content_type_deny=[re.compile(r"^application/x-fedc-test$")],
+        )
+
     async def test_wrong_content_type(self):
-        with self.assertRaises(CheckerError):
+        # Because so many web servers serve source code archives with an incorrect
+        # Content-Type, we ignore the Content-Type header and instead use
+        # (lib)magic to sniff the content type from the response data.
+        #
+        # However, we also have to deal with the problem that SourceForge
+        # sometimes returns a 200 OK response with an HTML error page rather
+        # than the source code archive we requested. So
+        # get_extra_data_info_from_url() allows the caller to provide a
+        # denylist.
+        #
+        # This test case is testing 2 things. The server returns a JSON body
+        # but with application/gzip as the Content-Type. We test that:
+        #
+        # 1. The content type is sniffed from the data as application/json
+        # 2. As a result it is rejected
+        with self.assertRaises(CheckerFetchError):
             await get_extra_data_info_from_url(
-                url=f"https://httpbin.org/response-headers?Content-Type={self._CONTENT_TYPE}",
+                url=f"https://httpbingo.org/response-headers?Content-Type=application/gzip",
                 session=self.http,
-                content_type_deny=[re.compile(f"^{self._CONTENT_TYPE}$")],
+                content_type_deny=[re.compile(r"^application/json$")],
             )
 
 
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -18,6 +18,9 @@ ignore_missing_imports = True
 [mypy-editorconfig.*]
 ignore_missing_imports = True
 
+[mypy-magic.*]
+ignore_missing_imports = True
+
 [mypy-semver.*]
 ignore_missing_imports = True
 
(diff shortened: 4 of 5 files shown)


gasinvein:
Unrelated CI failures should be fixed now, please rebase on current `master` and drop the second commit.

Alexander-Wilms:
Done

dependencies.apt.txt
```diff
@@ -22,6 +22,7 @@ python3-jsonschema
 python3-editorconfig
 python3-lxml
 python3-packaging
+libmagic1
```
gasinvein: We normally prefer distro-provided python module packages.
```suggestion
python3-magic
```

src/lib/utils.py
```diff
@@ -21,6 +21,7 @@
 import datetime as dt
 import json
 import logging
+import magic
```
gasinvein: 3rd party imports should go after standard library imports (into the 2nd imports list).

src/lib/utils.py
```diff
@@ -109,26 +110,40 @@ async def get_extra_data_info_from_url(
...
-                f"Wrong content type '{content_type}' received from '{url}'"
+            log.debug(
+                f"Wrong http content type '{http_content_type}' received from '{url}'"
             )
 
+        first_chunk = True
         checksum = MultiHash()
         size = 0
         async for chunk in response.content.iter_chunked(HTTP_CHUNK_SIZE):
+            if first_chunk:
```
gasinvein, line 131: I think we should try to detect content type from response body only when we previously got unexpected one in http header.

tests/test_util.py
```diff
@@ -222,12 +221,19 @@ async def asyncSetUp(self):
     async def asyncTearDown(self):
         await self.http.close()
 
+    async def test_correct_content_type(self):
+        await get_extra_data_info_from_url(
+            url=f"https://ftp.gnu.org/gnu/gzip/gzip-1.12.tar.gz",
```
gasinvein, line 227: For a software distribution site, wrong content type is clearly a bug that can (and should) be fixed some day and break out test. Instead, use some URL that reliably returns invalid content type (e.g. `https://httpbingo.org/response-headers?Content-Type=application/x-tar`), and verify that the new check raises an error.

gasinvein (changes requested):

src/lib/utils.py
Alexander-Wilms: Should I sort the imports?

grafik

src/lib/utils.py
Alexander-Wilms: You mean if this condition in line 122 is false?

            http_content_type is not None
            and content_type_deny is not None
            and any(r.match(http_content_type) for r in content_type_deny)

src/lib/utils.py
gasinvein: No, we don't use automatic sorter currently. Just put `magic` to the end of 3rd party modules.

tests/test_util.py
Alexander-Wilms: I think detection via magic number doesn't work with httpbingo:

alexander@PC:~/Downloads> file gzip-1.12.tar.gz 
gzip-1.12.tar.gz: gzip compressed data, max compression, from Unix, original size modulo 2^32 6041600
alexander@PC:~/Downloads> file response-headers\?Content-Type\=application%2Fx-tar 
response-headers?Content-Type=application%2Fx-tar: JSON text data

In the test case I'm checking against an incorrect mime type instead.

src/lib/utils.py
gasinvein: Yes, so that the logic becomes something like as follows: ff content type from headers matched denylist, then try to determine content type from the first chunk of data (and if it matched again, raise an error).
I guess you can put the denylist match check to a nested function like
```python
async def get_extra_data_info_from_url(...):
    def content_type_rejected(content_type: t.Optional[str]) -> bool:
        return (
            content_type is not None
            and content_type_deny is not None
            and any(r.match(content_type) for r in content_type_deny)
        )
    ...
```

tests/test_util.py
Alexander-Wilms: Ah, now I understand what you meant. That the httpbingo response has a mismatch between the actual content type as determined by `file` and the `content-type` field in the http header of the response, right?

Alexander-Wilms: In urlchecker.py, get_extra_data_info_from_url() is called without the content_type_deny parameter, and application/gzip is not among the WRONG_CONTENT_TYPES_ in src/lib/init.py, so content_type_rejected() would return false for https://httpbingo.org/response-headers?Content-Type=application/gzip and the magic number wouldn't be checked

Aren't these two disparate aspects? 1) getting the content type. This has in the past been done by checking the http response header. With my MR, it'd be determined using the magic number 2) rejecting URLs if the content type matches one of those in the deny list.
So I'd imagine the proper way would be to always get the content_type from the magic number of the first chunk and cancel the download if content_type_rejected(actual_content_type) returns true

Alexander-Wilms:
@wjt Could you review and merge?

tests/test_util.py
```diff
@@ -222,12 +222,19 @@ async def asyncSetUp(self):
...
+            session=self.http,
+            content_type_deny=[re.compile(f"^{self._CONTENT_TYPE}$")],
```
wjt: I don't think the class-level constant adds anything here – why not inline it?

```suggestion
            content_type_deny=[re.compile(r"^application/x-fedc-test$")],
```

tests/test_util.py
```diff
@@ -222,12 +222,19 @@ async def asyncSetUp(self):
...
+
     async def test_wrong_content_type(self):
-        with self.assertRaises(CheckerError):
+        with self.assertRaises(CheckerFetchError):
             await get_extra_data_info_from_url(
-                url=f"https://httpbin.org/response-headers?Content-Type={self._CONTENT_TYPE}",
+                url=f"https://httpbingo.org/response-headers?Content-Type=application/gzip",
                 session=self.http,
-                content_type_deny=[re.compile(f"^{self._CONTENT_TYPE}$")],
+                content_type_deny=[re.compile(f"^application/json$")],
```
wjt: Shouldn't be an f-string.

```suggestion
                content_type_deny=[re.compile(r"^application/json$")],
```

tests/test_util.py
```diff
@@ -222,12 +222,19 @@ async def asyncSetUp(self):
...
         await self.http.close()
 
+    async def test_correct_content_type(self):
+        await get_extra_data_info_from_url(
+            url=f"https://ftp.gnu.org/gnu/gzip/gzip-1.12.tar.gz",
+            session=self.http,
+            content_type_deny=[re.compile(f"^{self._CONTENT_TYPE}$")],
+        )
+
     async def test_wrong_content_type(self):
```
wjt, line 232: It's kind of weird that the only test of the new behaviour – ignoring the `Content-Type` header and instead guessing from the first chunk of data – is here, in a kind of double-negative way.

I had to refresh my memory of what this `content_type_deny` feature is for, and it's for Sourceforge which returns HTML pages with a 200 OK status code rather than source code archives in some scenarios.

OK, I guess this works. But can we have a comment explaining it, something like  (line-wrapped as appropriate):

```suggestion
    async def test_wrong_content_type(self):
        # Because so many web servers serve source code archives with an incorrect
        # Content-Type, we ignore the Content-Type header and instead use
        # (lib)magic to sniff the content type from the response data.
        #
        # However, we also have to deal with the problem that SourceForge
        # sometimes returns a 200 OK response with an HTML error page rather
        # than the source code archive we requested. So
        # get_extra_data_info_from_url() allows the caller to provide a
        # denylist.
        #
        # This test case is testing 2 things. The server returns a JSON body
        # but with application/gzip as the Content-Type. We test that:
        #
        # 1. The content type is sniffed from the data as application/json
        # 2. As a result it is rejected
```

wjt (changes requested):

tests/test_util.py
Alexander-Wilms: Yes, it's a bit confusing. I've added your explanation.

wjt (approved):

wjt:
Thank you for your persistence!

Alexander-Wilms:
Sure thing

PR #341, opened 2023-01-14
update jetbrains checker to add support for arm64 builds

binaryfields:
JetBrains publishes separate builds for Linux ARM64. This adds support to jetbrains checker to include arch property and grab the appropriate download node from the call to fetch available builds.

### DIFF
--- a/src/checkers/jetbrainschecker.py
+++ b/src/checkers/jetbrainschecker.py
@@ -7,6 +7,11 @@ from ..lib.checkers import Checker
 
 log = logging.getLogger(__name__)
 
+DOWNLOAD_NODE_SUFFIX = {
+    "x86_64": "",
+    "aarch64": "ARM64",
+}
+
 
 class JetBrainsChecker(Checker):
     CHECKER_DATA_TYPE = "jetbrains"
@@ -33,9 +38,8 @@ class JetBrainsChecker(Checker):
             result = await response.json()
             data = result[code][0]
 
-        release = data["downloads"][
-            self._get_download_node_name(external_data.arches[0])
-        ]
+        arch = external_data.arches[0]
+        release = data["downloads"][f"linux{DOWNLOAD_NODE_SUFFIX[arch]}"]
 
         async with self.session.get(release["checksumLink"]) as response:
             result = await response.text()
@@ -50,11 +54,3 @@ class JetBrainsChecker(Checker):
         )
 
         external_data.set_new_version(new_version)
-
-    @staticmethod
-    def _get_download_node_name(arch: str) -> str:
-        mapping = {
-            "x86_64": "",
-            "aarch64": "ARM64",
-        }
-        return f"linux{mapping[arch]}"
--- a/src/checkers/jetbrainschecker.py
+++ b/src/checkers/jetbrainschecker.py
@@ -16,7 +16,6 @@ class JetBrainsChecker(Checker):
             "code": {"type": "string"},
             # TODO: add enum here
             "release-type": {"type": "string"},
-            "arch": {"type": "string"},
         },
         "required": ["code"],
     }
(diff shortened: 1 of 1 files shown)


binaryfields:
@gasinvein Would you be able to take a look at this PR? 

The test failures seem related to tooling changes  for type check. 

For test cases, I'm not sure why expected failure is 2 instead of 3, but I assume is probably not related to this PR.

src/checkers/jetbrainschecker.py
```diff
@@ -33,7 +34,12 @@ async def check(self, external_data: ExternalBase):
             result = await response.json()
             data = result[code][0]
 
-        release = data["downloads"]["linux"]
+        arch = (
+            external_data.checker_data["arch"]
```
gasinvein: `ExternalData` objects already have `arches` property on them, you can access it as `external_data.arches[0]`, and it also defaults to `x86_64`. No need to re-introduce similar options just for this specific checker.

src/checkers/jetbrainschecker.py
```diff
@@ -48,3 +54,11 @@ async def check(self, external_data: ExternalBase):
         )
 
         external_data.set_new_version(new_version)
+
+    @staticmethod
+    def _get_download_node_name(arch: str) -> str:
```
gasinvein: I think this static mapping should be a class property.

gasinvein (changes requested):
I've fixed the unrelated CI failures, please rebase on current `master`.

src/checkers/jetbrainschecker.py
binaryfields: Good to know. I'll drop the property and wire it up as you suggested.

src/checkers/jetbrainschecker.py
binaryfields: i don't think property would work as I need to access external data argument passed to check method. I think this would still be a function.

gasinvein: Your initial submission (before commit 5c96449a30c659bec15f1b631d36a3668e7dcb47) was mostly correct, you should've just accessed the class property inside a method as `self.DOWNLOAD_NODE_SUFFIX`, not just `DOWNLOAD_NODE_SUFFIX`.

binaryfields: Got it. I'll swap it out.

binaryfields: I simplified the code and made download node suffix static.

binaryfields: Please let me know if you would like me to make any other changes. Thanks!

src/checkers/jetbrainschecker.py
binaryfields: Just to confirm, would external_data.arches[0] return the architecture of the currently built package variant? I'm unclear why this would have to be an array.

gasinvein: It refers to `only-arches` property of the source, which is an array of arches the source is used on, although the array usually contains only one item.

src/checkers/jetbrainschecker.py
```diff
@@ -7,6 +7,11 @@
 
 log = logging.getLogger(__name__)
 
+DOWNLOAD_NODE_SUFFIX = {
+    "x86_64": "",
+    "aarch64": "ARM64",
```
gasinvein:
```suggestion
_JB_ARCH_MAP = {
    "x86_64": "linux",
    "aarch64": "linuxARM64",
```

src/checkers/jetbrainschecker.py
```diff
@@ -33,7 +38,8 @@ async def check(self, external_data: ExternalBase):
             result = await response.json()
             data = result[code][0]
 
-        release = data["downloads"]["linux"]
+        arch = external_data.arches[0]
+        release = data["downloads"][f"linux{DOWNLOAD_NODE_SUFFIX[arch]}"]
```
gasinvein:
```suggestion
        release = data["downloads"][_JB_ARCH_MAP[external_data.arches[0]]]
```

gasinvein (commented):
I think it would be a bit cleaner this way:

src/checkers/jetbrainschecker.py
binaryfields: I pulled in your changes to simplify the code. Thank you for thanking the time to look it over and provide your feedback. Cheers!

gasinvein (approved):

gasinvein:
Thanks.

PR #353, opened 2023-03-04
checker: disable content type check in _get_json

hexchain:
It seems that `aiohttp.ClientResponse.json` refuses to decode if the response content type isn't `application/json`. To workaround bad APIs in the wild ([example](https://buildbot.flathub.org/#/builders/6/builds/26538)) that returns valid JSON with an incorrect content type, disable this check.

### DIFF
--- a/src/checkers/__init__.py
+++ b/src/checkers/__init__.py
@@ -4,6 +4,7 @@ import logging
 from distutils.version import LooseVersion
 from string import Template
 import datetime
+import json
 import re
 import typing as t
 import importlib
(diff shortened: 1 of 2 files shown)


src/checkers/__init__.py
```diff
@@ -120,7 +120,7 @@ async def _get_json(
                         return yaml.load(await response.read())
                     except ruamel.yaml.error.YAMLError as err:
                         raise CheckerQueryError("Failed to parse YAML") from err
-                return await response.json()
+                return await response.json(content_type=None)
```
gasinvein: We'll need to wrap this in a try/except block to capture and re-raise json decode errors, because `response.json()` won't raise `aiohttp.client.ContentTypeError` anymore (which was captured in the outer try/except block).
```suggestion
                try:
                    return await response.json(content_type=None)
                except json.JSONDecodeError as err:
                    raise CheckerQueryError from err
```

And, ideally, please add a test case to verify that a known non-json url causes a `CheckerQueryError`.

gasinvein (changes requested):

src/checkers/__init__.py
hexchain: All done.

gasinvein (approved):
Great, thanks!

hexchain:
How long would it take for this change to be available in Buildbot?

gasinvein:
Normally about 4 hours after the update was merged, just like with all other apps on Flathub. But I've just manually requested it to publish, should be there soon.

PR #356, opened 2023-03-09
main: Ensure git checkout is a safe directory

### LINKED ISSUE #352: Git can't commit because of "dubious ownership"
I'm the maintainer for https://github.com/flathub/org.signal.Signal and we recently introduced the beta branch of Signal. From my understanding, we need an own Github action for the beta branch.
This worked until yesterday.

I guess the fedc container updated the git version and is now running git in an environment which is deemed "unsafe" by git.
You can see from the action log: https://github.com/flathub/org.signal.Signal/actions/runs/4313990291/jobs/7526468080

```
[...]
fatal: detected dubious ownership in repository at '/github/workspace'
[...]
subprocess.CalledProcessError: Command '['git', 'branch', '--show-current']' returned non-zero exit status 128.
```

This issue is also tracked for the [actions/checkout #1169](https://github.com/actions/checkout/issues/1169). There seems to be no nice solution and only a workaround by running
```
git config --global --add safe.directory /github/workspace
```
*inside* the container before committing changes.

dbnicholson:
When running the checker in a container containing a recent git, it will often fail because the container user is not the same as the user owning the repository. Register the checkout as a safe directory if it isn't already.

I haven't tested this, but I'm pretty confident it will work.

Fixes: #352

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -27,6 +27,7 @@ import contextlib
 import json
 import logging
 import os
+from pathlib import Path
 import subprocess
 import sys
 import asyncio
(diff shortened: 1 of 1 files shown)


wjt (approved):
Lgtm!

PR #359, opened 2023-03-17
Versions of the format 0.x.y should be considered stable

### LINKED ISSUE #338: Script requests to update to an older version
In https://github.com/flathub/org.gnome.Evince/pull/110 the script is recommending to downgrade, which is odd unless there is something I am missing.

```
Update libgxps-0.3.2.tar.xz to 0.2.5
```

0.3.2 archive was uploaded on 2021-02-26 whereas 0.2.5 on 2017-02-25

* https://download.gnome.org/sources/libgxps/0.3/
* https://download.gnome.org/sources/libgxps/0.2/

### LINKED ISSUE #358: Incorrect downgrade of rest.
From this PR f-e-d-c want to downgrade rest

https://github.com/flathub/org.gnome.Photos/pull/46

However I can't figure out why not why was it just triggered. But 0.8.1 is definitely not more recent than 0.9.1, neither in date nor in order.

ManuelHu:
Fixes #338, fixes #358

This will consider all versions with a maj́or version of 0 as stable when using the gnome checker.

---

This is currently not covered by tests, as alleyoop only has one version branch 0.9.x: https://github.com/flathub/flatpak-external-data-checker/blob/890dd4b8d6fca69539ccdc5c53727f55716877fc/tests/org.gnome.baobab.json#L15-L18
this will not trigger the behaviour, as it will fallback to the single unstable version branch if no other stable version is available.

### DIFF
--- a/src/checkers/gnomechecker.py
+++ b/src/checkers/gnomechecker.py
@@ -32,7 +32,7 @@ def _is_stable(version: str) -> bool:
     major, minor = ver_list[:2]
     if set(ver_list[1:]) & {"alpha", "beta", "rc"}:
         return False
-    if int(major) < 40 and len(ver_list) > 2:
+    if int(major) > 0 and int(major) < 40 and len(ver_list) > 2:
         return (int(minor) % 2) == 0
     # XXX If we didn't see any indication that the version is a prerelease,
     # assume it's a normal (stable) release


gasinvein:
I've fixed the unrelated test failure (timeout on httpbin.org) and rebased your branch.

Although I'm not entirely sure this is correct. Doesn't Gnome use the odd-for-unstable version scheme for 0.x.y versions, too?

hfiguiere:
I think #374 would help a lot.

PR #372, opened 2023-06-10
Always create description tag

razzeee:
I think it makes sense to create these empty tags always. Even if it means there will be a few more bytes to download.

I hope this will motivate some more maintainers to actually fill out the description. 

Also it will be handy for me, cause I always need to look up the exact syntax whenever I get an update, as I usually forget it.

### DIFF
--- a/src/lib/appdata.py
+++ b/src/lib/appdata.py
@@ -83,6 +83,12 @@ def add_release(
     releases.insert(0, release)
     _fill_padding(release)
 
+    description = ElementTree.Element("description")
+    description.text = ''
+
+    release.append(description)
+    _fill_padding(description)
+
     tree.write(
         dst,
         # XXX: lxml uses single quotes for doctype line if generated with
--- a/src/lib/appdata.py
+++ b/src/lib/appdata.py
@@ -91,7 +91,8 @@ def add_release(
 
     # Indent the opening <description> tag one level
     # deeper than the <release> tag.
-    release.text = "\n" + ((releases.text[1::2]) * 3)
+    if releases.text:
+        release.text = "\n" + ((releases.text[1::2]) * 3)
 
     # Indent the closing </release> tag by the same amount as the opening
     # <release> tag (which we know to be the first child of <releases> since
--- a/src/lib/appdata.py
+++ b/src/lib/appdata.py
@@ -84,10 +84,20 @@ def add_release(
     _fill_padding(release)
 
     description = ElementTree.Element("description")
-    description.text = ''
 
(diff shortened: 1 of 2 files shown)


razzeee:
I'm also not sure how to get that one test to have the correct padding

wjt:
I briefly looked at the padding today. It's confusing and I can't quite remember why I felt it to be so important to preserve formatting in badly-formatted files. I think it was because if this automation was to be forced on 500+ apps, I wanted it to be as well-behaved as possible so people didn't hate its PRs.

Maybe the logic could be made less fancy and/or that test dropped. But it should be possible to get it to work. I'll try to get back into the mind of myself circa 2019 on a flight home from Greece..

wjt:
OK, the logic was rewritten since that flight home from Greece, but I kind of got my head around it and figured out something that works. Unfortunately I also discovered that the existing logic is broken in some edge cases, but I marked that new test case as xfail and moved on!

src/lib/appdata.py
```diff
@@ -83,6 +83,22 @@ def add_release(
...
 
+    description = ElementTree.Element("description")
+
+    # Give <description> a closing </description> rather than it being
+    # self-closing
+    description.text = ""
+
+    # Indent the opening <description> tag one level
+    # deeper than the <release> tag.
+    release.text = "\n" + ((releases.text[1::2]) * 3)
```
wjt: mypy objects to this because, while I know that `releases.text` is not `None`, mypy can't prove it.

Fair enough but it makes the problem harder to fix…

razzeee:
So how to proceed?

razzeee:
@wjt can we move this along?

wjt:
Sorry, I don't have time to work on this right now. Perhaps you could add an assertion to placate mypy?

razzeee:
I can certainly try, it's only about the test failing?

tests/test_appdata.py
```diff
@@ -46,13 +46,90 @@ def test_simple(self):
...
     <release version="1.2.3" date="2019-01-01"/>
   </releases>
 </component>
             """.strip(),
         )
 
+    @unittest.expectedFailure
+    def test_four_space_no_releases_element(self):
+        # FIXME: This ends up indenting <releases> correctly, but
+        # <release> and <description> incorrectly get the default 2-space
```
wjt, line 61: Oh I'd forgotten about this one. I think it's such an edge case that we can accept the failure for now.

src/lib/appdata.py
razzeee: I guess an `if` case might help? Pushed an approach.

PR #377, opened 2023-08-04
Report real URL for bad content types

wjt:
We believe that a KDE mirror is responsible for returning text/plain for source tarballs, but the raised exception only includes the original URL, not the result of following any redirects.

Report both, if they differ.

See #315


razzeee:
Seems sane to me

PR #392, opened 2023-11-05
Fix tests and add builddeps of python3-apt to Dockerfile

bbhtt:
The tests broke a bit after https://github.com/flathub/flatpak-external-data-checker/pull/391

Trying to fix the installation part of all pip modules at least.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,8 +4,11 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 ADD dependencies.apt.txt ./
 
+RUN printf "deb-src http://deb.debian.org/debian bullseye main\ndeb-src http://deb.debian.org/debian-security/ bullseye-security main\ndeb-src http://deb.debian.org/debian bullseye-updates main" >> /etc/apt/sources.list
+
 RUN apt-get update && \
     xargs apt-get install --no-install-recommends -y < dependencies.apt.txt && \
+    apt-get --no-install-recommends -y build-dep python3-apt && \
     apt-get clean && \
     rmdir /var/cache/apt/archives/partial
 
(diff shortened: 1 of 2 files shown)


bbhtt:
Better to install the entire build-essential metapackage I guess

bbhtt:
Ok it should be back to the state before https://github.com/flathub/flatpak-external-data-checker/pull/391

bbhtt:
Ok, all tests should be fixed now.

bbhtt:
The issue with httpbin-go is now fixed https://github.com/mccutchen/go-httpbin/pull/153, I think I can drop the last commit.

Dockerfile
```diff
@@ -4,8 +4,11 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 ADD dependencies.apt.txt ./
 
+RUN printf "deb-src http://deb.debian.org/debian bullseye main\ndeb-src http://deb.debian.org/debian-security/ bullseye-security main\ndeb-src http://deb.debian.org/debian bullseye-updates main" >> /etc/apt/sources.list
+
 RUN apt-get update && \
     xargs apt-get install --no-install-recommends -y < dependencies.apt.txt && \
+    apt-get --no-install-recommends -y build-dep python3-apt && \
```
wjt, line 11: It feels a bit weird to install the build-deps from apt, but then install the package via pip. I guess that's fine because it makes the `requirements.txt` usable on non-Debian distros.

Dockerfile
```diff
@@ -4,8 +4,11 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 ADD dependencies.apt.txt ./
 
+RUN printf "deb-src http://deb.debian.org/debian bullseye main\ndeb-src http://deb.debian.org/debian-security/ bullseye-security main\ndeb-src http://deb.debian.org/debian bullseye-updates main" >> /etc/apt/sources.list
```
wjt, line 7: I might try a patch to generate this based on the existing lines of `/etc/apt/sources.list` but this will work!

wjt (approved):
Thank you for looking into this!

Dockerfile
bbhtt: I noticed it too, that pip modules are getting installed twice one by apt then by pip.

Not sure what to do there.

PR #393, opened 2023-11-08
Update container image to Bookworm

wjt:


Dockerfile
```diff
@@ -1,30 +1,23 @@
...
     apt-get --no-install-recommends -y build-dep python3-apt && \
     apt-get clean && \
     rmdir /var/cache/apt/archives/partial
 
+# All requirements should be satisfied by dependencies.apt.txt – but feed it to
+# pip to keep us honest
 ADD requirements.txt ./
-
-RUN python3 -m pip install -r requirements.txt && \
+RUN sed -i 's/python-apt @ .*/python-apt/' requirements.txt && \
```
bbhtt, line 21: Maybe it'd be better to filter python-apt from requirements.txt instead, python3-apt from Debian repositories and the current source are both the same thing.

This replaces it with an outdated yanked version from pypi

wjt: No, it doesn't.

```
Requirement already satisfied: python-apt in /usr/lib/python3/dist-packages (from -r requirements.txt (line 9)) (2.6.0)
```

I couldn't find a way to make `pip` check that everything listed is installed without actually downloading anything, but that's what I actually wanted to do.

wjt: I found a way to do this.

wjt:
No wonder the tests are slow – testing the Chromium checker downloads a 3 GiB Chromium source tarball!

#394

Dockerfile
bbhtt: Ah yea the one from apt is installed first

PR #401, opened 2023-12-07
chromiumchecker: Update to the new Chromium Dash API

refi64:
The old one apparently went offline, oops.


refi64:
(test failure seems unrelated?)

wjt:
Test failure does seem unrelated.

a5a8ee61 introduced a test that 10 bytes of random data are fetched and then we just hope that they don't trip up the charset detector in aiohttp – seems a bad idea.

PR #403, opened 2023-12-11
Add support for max_line_length (EditorConfig)

### LINKED ISSUE #280: Prevent line truncation in YAML manifests
Hello,

In [this pull request](https://github.com/flathub/br.gov.fazenda.receita.irpf/pull/8), the bot unnecessarily truncates line 34.

I'm not sure if Flathub has a "max line-length" guideline, but nevertheless this can be properly configured using the `width` option of ruamel, e.g. at:

https://github.com/flathub/flatpak-external-data-checker/blob/88e46ff8ee3176b4fea752876304e4a249765462/src/lib/utils.py#L417-L422

We'd add:

```python
# allows lines to have at most one thousand characters
_yaml.width = 1000
```

Thanks!

AtomHare:
This PR adds support for the `max_line_length` following the [EditorConfig spec](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#supported-by-a-limited-number-of-editors). 
Directly follows #280 and linked to a recent PR in a repo I manage https://github.com/flathub/io.github.everestapi.Olympus/pull/2.

Fixes #280.

### DIFF
--- a/src/lib/utils.py
+++ b/src/lib/utils.py
@@ -548,13 +548,11 @@ def dump_manifest(contents: t.Dict, manifest_path: t.Union[Path, str]):
         indent = 4
 
     # Determine max line length preference
-    max_line_length: t.Optional[int]
-    if "max_line_length" in conf:
+    if max_line_length := conf.get("max_line_length"):
         try:
-            max_line_length = int(conf.get("max_line_length"))
-            _yaml.width = max_line_length  # type: ignore # See https://sourceforge.net/p/ruamel-yaml/tickets/322/
+            _yaml.width = int(max_line_length)  # type: ignore # See https://sourceforge.net/p/ruamel-yaml/tickets/322/
         except ValueError:
-            log.error("max_line_length EditorConfig variable is not valid")
+            log.warning("Ignoring invalid max_line_length %r", max_line_length)
 
     # Determine trailing newline preference
     newline: t.Optional[bool]
--- a/src/lib/utils.py
+++ b/src/lib/utils.py
@@ -548,6 +548,7 @@ def dump_manifest(contents: t.Dict, manifest_path: t.Union[Path, str]):
         indent = 4
 
     # Determine max line length preference
+    max_line_length: t.Optional[int]
     if "max_line_length" in conf:
         try:
             max_line_length = int(conf.get("max_line_length"))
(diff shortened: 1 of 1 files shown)


src/lib/utils.py
```diff
@@ -547,6 +547,15 @@ def dump_manifest(contents: t.Dict, manifest_path: t.Union[Path, str]):
     else:
         indent = 4
 
+    # Determine max line length preference
+    max_line_length: t.Optional[int]
+    if "max_line_length" in conf:
+        try:
+            max_line_length = int(conf.get("max_line_length"))
+            _yaml.width = max_line_length
```
bbhtt: I think you need to ignore the type check here, ruamel-yaml sets it as None https://sourceforge.net/p/ruamel-yaml/tickets/322/ which is a bug on their part.

AtomHare: Nice catch, edited to ignore the mypy check

src/lib/utils.py
```diff
@@ -547,6 +547,15 @@ def dump_manifest(contents: t.Dict, manifest_path: t.Union[Path, str]):
...
         indent = 4
 
+    # Determine max line length preference
+    max_line_length: t.Optional[int]
+    if "max_line_length" in conf:
+        try:
+            max_line_length = int(conf.get("max_line_length"))
+            _yaml.width = max_line_length  # type: ignore # See https://sourceforge.net/p/ruamel-yaml/tickets/322/
+        except ValueError:
+            log.error("max_line_length EditorConfig variable is not valid")
```
wjt: Error seems a bit severe. I'd make this a warning. And mypy annotations aside I think something like this would be clearer:

```suggestion
    if max_line_length := conf.get("max_line_length"):
        try:
            _yaml.width = int(max_line_length)  # type: ignore # See https://sourceforge.net/p/ruamel-yaml/tickets/322/
        except ValueError:
            log.warning("Ignoring invalid max_line_length %r", max_line_length)
```

AtomHare: You are right, raising an error might be too much, a simple warning is ideal. 
Regarding your suggestions it looks good to me but do we want to drop `max_line_length: t.Optional[int]` ? I placed it in order to follow the same syntax as the others EditorConfig parameters

wjt: I personally don't think it adds anything because we are doing our own runtime type-check, and `mypy` can surely infer that `int(max_line_length)` is of type `int`.

AtomHare: Corrected the code following your recommandations

wjt (approved):

wjt:
Thank you!

PR #414, opened 2024-04-25
Don't clear environment variables

### LINKED ISSUE #413: GIT_AUTHOR_NAME and GIT_AUTHOR_EMAIL filtered from environment
The documented way to run the checker and get the desired git author information is to set the `GIT_AUTHOR_NAME` and `GIT_AUTHOR_EMAIL` environment variables. Unfortunately, these are [filtered out](https://github.com/flathub-infra/flatpak-external-data-checker/commit/16a3be10) because they contain the phrase `auth`.

I think this is also why the documentation says to set `EMAIL`. It shouldn't be needed except that `GIT_AUTHOR_EMAIL` doesn't get through. In endless-key-flatpak, our [commits](https://github.com/endlessm/endless-key-flatpak/commit/acc041c395febe01114ef93f05d7e462576b2f4a) are being authored with `root <os@endlessos.org>`. In other words, we get the email address because we also set `EMAIL` in the workflow, but we lose our desired `GIT_AUTHOR_NAME`.

Not sure the best way to handle it. The sensitive variable filtering seems useful, and it does make sense that environment variables containing `auth` are likely to be one of those. On the other hand, why is the environment being filtered? Because it's dumped to the output? I would think some environment variable containing credentials might actually be required to run f-e-d-c depending on what the checkers do or what the git push does. @gasinvein?

dbnicholson:
See the first commit for discussion, but basically it can't be done correctly like this and it was causing the `GIT_AUTHOR_*` environment variables to be unset when making commits. A couple tests are added to make sure that doesn't happen again.

Fixes: #413

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,6 @@ jobs:
           # email sets "github-actions[bot]" as commit author, see https://github.community/t/github-actions-bot-email-address/17204/6
           GIT_AUTHOR_EMAIL: 41898282+github-actions[bot]@users.noreply.github.com
           GIT_COMMITTER_EMAIL: 41898282+github-actions[bot]@users.noreply.github.com
-          EMAIL: 41898282+github-actions[bot]@users.noreply.github.com
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           args: --update --never-fork $PATH_TO_MANIFEST # e.g. com.organization.myapp.json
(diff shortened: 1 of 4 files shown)


wjt (approved):
LGTM – presumably if you rebase the test failures will disappear since your branch at #415 was merged?

dbnicholson:
> LGTM – presumably if you rebase the test failures will disappear since your branch at #415 was merged?

It should. Rebased now.

dbnicholson:
> > LGTM – presumably if you rebase the test failures will disappear since your branch at #415 was merged?
> 
> It should. Rebased now.

And yet `flake8` has done its job and found and unused import.

PR #419, opened 2024-05-02
fixed the external data checker from stalling when git tag pattern is incorrect, this should fix the github actions run

### LINKED ISSUE #418: I found why the external data checker fails on github actions, I figured out the solution
When an app uses git as the source, and uses a tag pattern that fails this assert:
https://github.com/flathub-infra/flatpak-external-data-checker/blob/6ab69bcbcc35507104198c17895d36b4113f204a/src/checkers/gitchecker.py#L101 , such as this app currently: https://github.com/flathub/com.modrinth.ModrinthApp/blob/1d954cc7c91841d020fe4c860593789a16ca62c6/com.modrinth.ModrinthApp.yml#L108 , the external data checker should skip this app entirely. 
If the checker is run from the source  code directly, this works. 
However, if the checker is run from the Flatpak or the docker container the checker stalls for the rest of the run. On GitHub actions it is getting cancelled by GitHub after 6 hours.
I have submitted a pull request to the offending app to fix the regex, but it seems this part of the external-data-checker should be fixed. I am not sure how though.

tobiasmicheler:
I recently opened a related issue:
This closes https://github.com/flathub-infra/flatpak-external-data-checker/issues/418
In this issue I said, that I don't know how to fix this problem. However, I figured it out.

### DIFF
--- a/src/checkers/__init__.py
+++ b/src/checkers/__init__.py
@@ -180,6 +180,29 @@ class Checker:
         except (KeyError, ValueError) as err:
             raise CheckerMetadataError("Error substituting template") from err
 
+    @classmethod
+    def _get_pattern(
+        cls,
+        checker_data: t.Dict,
+        pattern_name: str,
+        expected_groups: int = 1,
+    ) -> t.Optional[re.Pattern]:
+        try:
+            pattern_str = checker_data[pattern_name]
+        except KeyError:
+            return None
+
+        try:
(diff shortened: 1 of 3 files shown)


wjt:
Thanks, this does fix the problem. Nice work tracking this down!

I'll amend your commit slightly for style.

What I am trying to work out is: why does the assertion failure hang the asyncio loop forever?

PR #423, opened 2024-05-26
externaldata: Error cleanly when source is missing required checksum

bbhtt:
See https://github.com/flathub-infra/flatpak-external-data-checker/issues/421

### DIFF
--- a/src/lib/externaldata.py
+++ b/src/lib/externaldata.py
@@ -248,6 +248,19 @@ class ExternalBase(BuilderSource):
         if not source.get("url"):
             raise SourceUnsupported('Data is not external: no "url" property')
 
+        if (
+            source.get("type") == ExternalBase.Type.EXTRA_DATA
+            and "sha256" not in source
+        ):
+            raise SourceLoadError("Extra data is missing SHA256 checksum")
+
+        if source.get("type") in (
+            ExternalBase.Type.FILE,
+            ExternalBase.Type.ARCHIVE,
+        ):
+            if not (source.keys() & {"md5", "sha1", "sha256", "sha512"}):
+                raise SourceLoadError("Data is missing checksum")
+
         # FIXME: https://github.com/python/mypy/issues/9282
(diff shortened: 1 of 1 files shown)


src/lib/externaldata.py
```diff
@@ -248,6 +248,23 @@ def from_source(
...
+
+            if source.get("type") in (
+                ExternalBase.Type.FILE,
+                ExternalBase.Type.ARCHIVE,
+            ):
+                if not any(
+                    k in source.keys() for k in ("md5", "sha1", "sha256", "sha512")
+                ):
+                    raise SourceLoadError("Data is missing checksum")
+
```
wjt: The `if source.get("url")` guard is unneeded because it was checked right above.

You can also simplify the checksum checks:

```suggestion
        if (
            source.get("type") == ExternalBase.Type.EXTRA_DATA
            and "sha256" not in source
        ):
            raise SourceLoadError("Extra data is missing SHA256 checksum")

        if source.get("type") in (
            ExternalBase.Type.FILE,
            ExternalBase.Type.ARCHIVE,
        ):
            if not (source.keys() & {"md5", "sha1", "sha256", "sha512"}):
                raise SourceLoadError("Data is missing checksum")
```

wjt (changes requested):

src/lib/externaldata.py
bbhtt: I don't want it to trigger when path was used, that does not require a checksum.

wjt: The check above:

```python3
        if not source.get("url"):
            raise SourceUnsupported('Data is not external: no "url" property')
```

guarantees that `if source.get("url")` is true if reached, so the additional check is unnecessary. Or am I missing something?

bbhtt: Ah no you're right, I didn't see the line above.

bbhtt:
release-monitoring.org went down.

wjt (approved):

PR #425, opened 2024-05-31
manifest: Skip updating appdata if last source is not external

### LINKED ISSUE #346: f-e-d-c crashes on updating AppStream
f-e-d-c currently crashes on updating the [6.4 branch of com.riverbankcomputing.PyQt.BaseApp](https://github.com/flathub/com.riverbankcomputing.PyQt.BaseApp/tree/branch/6.4). 

```
INFO    src.manifest: Updating /github/workspace/pyqt/../pyqt-builder/pyqt-builder.json
INFO    src.manifest: Updating /github/workspace/pyqt/../pyqt-sip/pyqt-sip.json
INFO    src.manifest: Updating /github/workspace/pyqt/pyqt.json
INFO    src.manifest: Preparing to update appdata /github/workspace/com.riverbankcomputing.PyQt.BaseApp.metainfo.xml
  File "/app/flatpak-external-data-checker", line 30, in <module>
    main()
  File "/app/src/main.py", line 457, in main
    return future.result()
  File "/app/src/main.py", line 425, in run_with_args
    changes = manifest_checker.update_manifests()
  File "/app/src/manifest.py", line 546, in update_manifests
    self._update_appdata()
  File "/app/src/manifest.py", line 492, in _update_appdata
    assert selected_data is not None
AssertionError
```

The full log can be found in the [Action](https://github.com/flathub/com.riverbankcomputing.PyQt.BaseApp/actions/runs/4041799929/jobs/6948751257).

bbhtt:
Closes https://github.com/flathub-infra/flatpak-external-data-checker/issues/346

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -486,10 +486,19 @@ class ManifestChecker:
             elif data.source_path == self._root_manifest_path:
                 selected_data = data
         else:
-            # Guess that the last external source in the root manifest is the one
-            # corresponding to the main application bundle.
-            assert selected_data is not None
-            log.warning("Guessed upstream source: %s", selected_data)
+            if selected_data is not None:
+                # Guess that the last external source in the root manifest is the one
+                # corresponding to the main application bundle.
+                log.warning("Guessed last source as main source: %s", selected_data)
+            else:
+                log.error(
+                    (
+                        "No main source configured and no external source in "
+                        "%s. Not updating appdata"
+                    ),
+                    self._root_manifest_path,
+                )
+                return
 
         last_update = selected_data.new_version
 


src/manifest.py
```diff
@@ -486,10 +486,15 @@ def _update_appdata(self):
...
-            # corresponding to the main application bundle.
-            assert selected_data is not None
-            log.warning("Guessed upstream source: %s", selected_data)
+            if selected_data is not None:
+                # Guess that the last external source in the root manifest is the one
+                # corresponding to the main application bundle.
+                log.warning("Guessed last source as main source: %s", selected_data)
+            else:
+                log.error(
+                    "The last source is not external. Appdata will not be updated"
```
wjt: Looking at https://github.com/flathub/com.riverbankcomputing.PyQt.BaseApp/blob/branch/6.4/com.riverbankcomputing.PyQt.BaseApp.yaml I don't think this error message is correct.

The condition is that there is no external source in the root manifest.

```suggestion
                log.warning(
                    "No main source configured; and no external source in %s. Not updating appdata.",
                    self._root_manifest_path,
```

bbhtt: Seems fine, will change it.

bbhtt: Actually saying there is no external source seems wrong? The manifest can have external source but if there is no `is-main-source` and the last one is not external, it will fail?

"No external source" would mean there was no external source anywhere.

bbhtt: slightly changed it.

wjt: > The manifest can have external source but if there is no is-main-source and the last one is not external, it will fail?

The loop runs across all external sources in order of appearance. If a source has the main-source property, `selected_data` is set and the loop terminated early. Otherwise, if the source is in the root manifest, `selected_data` is updated. Finally, if the loop has not terminated early, this block runs.

So if this block is reached, if there is 1 or more external source in the root manifest, `selected_data` will be non-`None`; otherwise it will be `None`.

So I think it is correct to say that `selected_data` will be `None` in this block only if there are *no* external sources in the root manifest.

bbhtt: Hm ok I thought the bug happened even with an external source early (with no is-main-source) and a non external one as the last one.

I'll change it back then.

bbhtt:
@wjt this one is ready too I think?

PR #427, opened 2024-06-01
main: Be more resistant to git commit failures

bbhtt:


src/main.py
```diff
@@ -199,7 +200,37 @@ def commit_changes(changes: t.List[str]) -> CommittedChanges:
...
+
+    retry_commit = False
+    try:
+        check_call(["git", "commit", "-am", message])
+    except subprocess.CalledProcessError:
+        retry_commit = True
+        pass
+
+    if retry_commit:
+        log.warning("Committing failed. Falling back to a sanitised config")
```
wjt, line 212: What's the real-world case you're trying to fix here?

bbhtt: See https://github.com/flathub-infra/flatpak-external-data-checker/pull/426#issue-2328988033 and https://github.com/flathub-infra/flatpak-external-data-checker/pull/426#issuecomment-2143363912

Have the git config in `~/.config/git/config` file, try `--update` with the flatpak it fails. If you allow access `--filesystem=xdg-config/git`, it may fail on not having access to gpg folder, no pinenetry, no gpg socket and multiple other errors that can arise from personal `gitconfig`.

bbhtt: In those cases I think it is good to commit with a fallback so that uncommitted changes do not get lost

src/main.py
```diff
@@ -197,8 +198,39 @@ def commit_changes(changes: t.List[str]) -> CommittedChanges:
...
+                f"user.email={git_email}",
+                "commit",
+                "--no-verify",
+                "--no-gpg-sign",
+                "-am",
+                message,
+            ],
+            check=True,
+            env=env,
+        )
```
wjt, line 233: The point of the `check_call()` wrapper is to log the command being executed, which can be very useful in understanding what the program is doing. I know `subprocess.run()` is the more modern replacement and that `check_call()` is not used everywhere. I guess it is OK to not log the command line. It does have a computed parameter…

wjt (approved):
Personally I think it's fine for Git to fail if it is misconfigured. No data will be lost because it's all in the working copy. But I don't feel too strongly about this. If it's useful, OK.

bbhtt:
> Personally I think it's fine for Git to fail if it is misconfigured. 

It's more like - using git from sandbox while sharing a config with host is going to cause issues.

PR #428, opened 2024-06-01
manifest: Add module name of changed data in commit message

bbhtt:


bbhtt:
`name` is a required property, and is checked by jsonschema. It can be None when someone sets it to empty like `- name: `, in that case, they are on their own.

I don't think I need to handle the `None` case separately.

wjt:
> `name` is a required property, and is checked by jsonschema.

Where do you see this check? I don't see any uses of `jsonschema` to check the manifest as a whole, only the `x-checker-data` blocks on a per-checker basis.

src/manifest.py
```diff
@@ -452,11 +452,11 @@ def _update_manifest(self, path, datas, changes):
 
             data.update()
             if data.new_version.version is not None:
-                message = "Update {} to {}".format(
-                    data.filename, data.new_version.version
+                message = "{}: Update {} to {}".format(
+                    data.module, data.filename, data.new_version.version
```
wjt, line 456: I kind of see the argument that if people want these messages not to be ugly then they can name their modules, but it is kind of ugly to put `None: ` here!

bbhtt: It won't be `none`, name is a required property by flatpak-builder.

wjt: But you had to update some of the tests to expect the string `None: ` here. Does that mean that the manifests used by those tests would be rejected by f-b?

wjt: OK, I see your point: the name is [required by the schema](https://github.com/flatpak/flatpak-builder/blob/61e3a2e0dc0b24f9023648d11af4d0d9285ca10f/data/flatpak-manifest.schema.json#L494-L496) and, though flatpak-builder doesn't actually validate the schema (...) there is an [ad-hoc check that the name is non-NULL](https://github.com/flatpak/flatpak-builder/blob/61e3a2e0dc0b24f9023648d11af4d0d9285ca10f/src/builder-manifest.c#L254-L263).

gasinvein:
> > `name` is a required property, and is checked by jsonschema.
> 
> Where do you see this check? I don't see any uses of `jsonschema` to check the manifest as a whole, only the `x-checker-data` blocks on a per-checker basis.

I believe f-e-d-c will fail to load a module without the `name` property (with an unhandled `KeyError` exception) here

https://github.com/flathub-infra/flatpak-external-data-checker/blob/bf14aa11b4bc50e3b7ba54e74009f1f6a5212454/src/lib/externaldata.py#L73-L84

src/manifest.py
bbhtt: > But you had to update some of the tests to expect the string None: here

Yea those are partial jsons. I tried to fix them by adding name but flake8 complains too much about indentations and line lengths.

PR #429, opened 2024-06-04
main: Set module names as title

bbhtt:
Choose the "best" title with most module names while making sure it does
not cross 70 characters. The limit is to prevent too much verbosity in
commit messages and PR titles when a large set of modules are getting
updated.

If it crosses 70 characters, start deducting module names and add the
number of remaining updated modules.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -177,15 +177,39 @@ class CommittedChanges(t.NamedTuple):
     base_branch: t.Optional[str]
 
 
+def commit_message(changes: t.List[str]) -> str:
+    assert len(changes) >= 1
+
+    if len(changes) == 1:
+        return changes[0]
+
+    module_names = list(dict.fromkeys(list(i.split(":", 1)[0] for i in changes)))
+    for i in reversed(range(2, len(module_names) + 1)):
+        xs = module_names[: i - 1]
+        y = module_names[i - 1]
+        zs = module_names[i:]
+
+        if zs:
+            tail = f" and {len(zs)} more modules"
+            xs.append(y)
+        else:
+            tail = f" and {y} modules"
+
+        subject = "Update " + ", ".join(xs) + tail
+        if len(subject) <= 70:
+            return subject
+
+    return f"Update {len(module_names)} modules"
+
+
 def commit_changes(changes: t.List[str]) -> CommittedChanges:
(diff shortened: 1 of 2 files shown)


bbhtt:
56 is the max characters Github can hold in the title in a single line. The commit message is probably 70 or 80.

I'd like to keep some form of "backoff" parameter to prevent ugly PR titles.

I convert it to set to avoid duplicate module names see eg. https://github.com/flathub/org.kde.kdenlive/pull/395

wjt:
Something like this, but preserving the order of `changes`?

```python3
def build_commit_message(changes: t.List[str]) -> str:
    assert len(changes) >= 1

    if len(changes) == 1:
        return changes[0]

    module_names = list({i.split(":", 1)[0] for i in changes})
    for i in reversed(range(2, len(module_names) + 1)):
        print(i)
        xs = module_names[:i - 1]
        y = module_names[i - 1]
        zs = module_names[i:]

        if zs:
            tail = f" and {len(zs)} more"
            xs.append(y)
        else:
            tail = f" and {y}"

        subject = "Update " + ", ".join(xs) + tail
        if len(subject) <= 56:
            return subject

    return f"Update {len(changes)} modules"
```

bbhtt:
If you want to preserve the order in the list only there might be some shorter way I'll check soon.

> list({i.split(":", 1)[0] for i in changes})

This doesn't preserve order of changes

bbhtt:
>     module_names = list({i.split(":", 1)[0] for i in changes})

This won't preserve order of changes

bbhtt:
Pushed something that will preserve order. The rest of the formatting I don't have opinions about. But if you can give me it as a patch I can preserve authorship.

bbhtt:
Also how about adding `Update: a, b and n more *modules*` ? As these aren't sources necessarily a single module can have multiple sources with checker data associated.

bbhtt:
> But if you can give me it as a patch I can preserve authorship.

@wjt Would you prefer that I finish this?

wjt:
Sure, I don't mind about authorship!

bbhtt:
Done

bbhtt:
I changed the limit to 70 characters to allow more module names in the title.

wjt (approved):

PR #431, opened 2024-06-12
main: Report the module name when only one module changed

bbhtt:
Instead of falling back to len(module_names)


bbhtt:
See for example https://github.com/flathub/io.github.streetpea.Chiaki4deck/pull/39

The changes are unique but, the module name is not (we are removing duplicated module names). So there can be multiple changed sources but all belonging to one module.

PR #443, opened 2024-11-15
chromiumchecker: Use Gentoo-hosted tarball mirrors on 404

refi64:
Upstream keeps having issues in their CI pipelines, resulting in massively delayed tarball uploads. Gentoo has been doing custom uploads in the interim, so try to automatically fall back to those URLs on 404.

### DIFF
--- a/src/checkers/chromiumchecker.py
+++ b/src/checkers/chromiumchecker.py
@@ -60,12 +60,30 @@ class ChromiumComponent(Component):
         "https://commondatastorage.googleapis.com"
         "/chromium-browser-official/chromium-{version}.tar.xz"
     )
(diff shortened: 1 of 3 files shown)


wjt:
A little weird but I suppose it's better than not updating, and the process of building these tarballs seems quite transparent.

Excuse my ignorance: is it not feasible to build from Chromium Git?

refi64:
*looks at CI* I swear one of these days I will correctly run the tests before a PR *sign*.

> Excuse my ignorance: is it not feasible to build from Chromium Git?

The tricky part is that the tarballs aren't just from Git: they're a mix of Git sources combined with a bunch of gclient-retrieved things from GCS & co. It's possible to reproduce it externally, but it's pretty tricky.

PR #447, opened 2024-12-25
urlchecker: Extract version from deb for rotating url

zhsj:


zhsj:
I intended to use https://github.com/flathub/com.tencent.WeChat/blob/master/com.tencent.WeChat.yaml for test. But the download url is slow on GitHub runner infra, and causes the test to be flaky. So I changed it to the chrome url.

zhsj:
@wjt PTAL

bbhtt (approved):
Seems fine

wjt (approved):

PR #453, opened 2025-01-18
gnomechecker: Add enum to detect unstable versions

A6GibKm:
Adds an enum that allows specifying how unstable versions should be detected. Few modules are still being released and are not using the new GNOME version scheme.

This solves, for example libadwaita 1.7 being picked as unstable. 

cc: @alatiera @wjt

Supersedes: https://github.com/flathub-infra/flatpak-external-data-checker/pull/435

### DIFF
--- a/src/checkers/gnomechecker.py
+++ b/src/checkers/gnomechecker.py
@@ -2,6 +2,7 @@ import logging
 import re
 import typing as t
 
+from enum import StrEnum
 from yarl import URL
 
 from ..lib import OPERATORS_SCHEMA, NETWORK_ERRORS
@@ -14,6 +15,11 @@ from . import Checker
 log = logging.getLogger(__name__)
 
 GNOME_MIRROR = URL("https://download.gnome.org/")
+
+
+class VersionScheme(StrEnum):
+    DEFAULT = "default"
+    ODD_MINOR_IS_UNSTABLE = "odd-minor-is-unstable"
 
 
 def _parse_checksums(text: str) -> t.Dict[str, str]:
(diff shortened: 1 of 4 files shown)


A6GibKm:
I feel CI is broken, `enum.StrEnum` *does* exist https://docs.python.org/3/library/enum.html#enum.StrEnum since python 3.11.

bbhtt:
https://github.com/flathub-infra/flatpak-external-data-checker/pull/454 should allow you to use python 3.11 features.

A6GibKm:
Rebased and CI fixed. Thanks.

src/checkers/gnomechecker.py
```diff
@@ -16,6 +17,12 @@
 GNOME_MIRROR = URL("https://download.gnome.org/")
 
 
+class VersionScheme(StrEnum):
+    DEFAULT = "default"
+    ODD_MAJOR_IS_UNSTABLE = "odd-major-is-unstable"
```
wjt: Are there any projects that follow this scheme?

wjt (approved):
LGTM, with one question. I rebased this branch and added a patch to document the new key in the README.

src/checkers/gnomechecker.py
A6GibKm: I been trying to come up with an example. It might be better to leave this out.

A6GibKm: I removed that enum value, I am not sure if we have relevant examples.

A6GibKm: It can be added back later if needed.

A6GibKm:
@bbhtt the PR should be done, however CI is failing.

bbhtt:
That's a network blip.

bbhtt (approved):

PR #454, opened 2025-01-19
Update CI to use latest actions and images

bbhtt:


bbhtt:
the problem is it is running inside the debian container as root but the default cache path is owned by some else so cache is being disabled:

```
WARNING: The directory '/github/home/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag.
```

bbhtt:
Works now

image

wjt (approved):

Issue #456, opened 2025-01-23
Invalid archive member header on Guilded-Linux deb


bbhtt:
From https://github.com/flathub/flathub/actions/runs/12917235025

```
2025-01-22T23:15:22.6366831Z DEBUG   src.manifest: Source extra-data guilded/Guilded-Linux.deb: applying URLChecker
2025-01-22T23:15:26.0119132Z Traceback (most recent call last):
2025-01-22T23:15:26.0120503Z   File "/app/flatpak-external-data-checker", line 30, in <module>
2025-01-22T23:15:26.0121261Z     main()
2025-01-22T23:15:26.0134141Z   File "/app/src/main.py", line 564, in main
2025-01-22T23:15:26.0134783Z     outdated_num, errors_num, updated = asyncio.run(run_with_args(args))
2025-01-22T23:15:26.0135332Z                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-01-22T23:15:26.0135869Z   File "/usr/lib/python3.11/asyncio/runners.py", line 190, in run
2025-01-22T23:15:26.0136383Z     return runner.run(main)
2025-01-22T23:15:26.0136718Z            ^^^^^^^^^^^^^^^^
2025-01-22T23:15:26.0137155Z   File "/usr/lib/python3.11/asyncio/runners.py", line 118, in run
2025-01-22T23:15:26.0137688Z     return self._loop.run_until_complete(task)
2025-01-22T23:15:26.0138098Z            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-01-22T23:15:26.0138694Z   File "/usr/lib/python3.11/asyncio/base_events.py", line 653, in run_until_complete
2025-01-22T23:15:26.0139512Z     return future.result()
2025-01-22T23:15:26.0139848Z            ^^^^^^^^^^^^^^^
2025-01-22T23:15:26.0140232Z   File "/app/src/main.py", line 525, in run_with_args
2025-01-22T23:15:26.0140711Z     await manifest_checker.check(args.filter_type)
2025-01-22T23:15:26.0141194Z   File "/app/src/manifest.py", line 408, in check
2025-01-22T23:15:26.0141684Z     ext_data_checked = await asyncio.gather(*check_tasks)
2025-01-22T23:15:26.0142326Z                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #457, opened 2025-01-23
utils: flush tempfile after downloading from url

zhsj:
Updates: #456

### DIFF
--- a/src/lib/utils.py
+++ b/src/lib/utils.py
@@ -175,6 +175,9 @@ async def get_extra_data_info_from_url(
             if dest_io is not None:
                 dest_io.write(chunk)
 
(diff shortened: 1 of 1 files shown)


zhsj:
While I'm not sure if this can fix #456 but the missing flush may cause other process reading incomplete content.

bbhtt:
This seems to be correct but I'm having a hard time reproducing #456 again locally. But it started affecting tests too https://github.com/flathub-infra/flatpak-external-data-checker/actions/runs/14971551632 

Let's merge this as a tentative fix. I configured the tests to run on a schedule on my fork with this fix to see if the error happens again.

PR #463, opened 2025-02-24
utils: Preserve yaml header if present in source manifest

### LINKED ISSUE #462: Flathub bot strips YAML header
Hey, 

When the flathub bot updates the dependencies, it strips the YAML header, which trips my Yaml linter. Is there something I can do ?

Happened here https://github.com/flathub/com.play0ad.zeroad/pull/149/commits so I had to update the commit manually.

bbhtt:
Fixes: #462

### DIFF
--- a/src/lib/utils.py
+++ b/src/lib/utils.py
@@ -512,7 +512,10 @@ _yaml.indent(mapping=2, sequence=4, offset=2)
 def read_yaml_manifest(manifest_path: Path):
     """Read a YAML manifest from 'manifest_path'."""
     with manifest_path.open("r") as f:
-        return _yaml.load(f)
+        has_yaml_header = (_ := f.read(3)) == "---"
+        f.seek(0)
+        data = _yaml.load(f)
+    return data, has_yaml_header
 
 
 def read_manifest(manifest_path: t.Union[Path, str]):
@@ -524,7 +527,9 @@ def read_manifest(manifest_path: t.Union[Path, str]):
         return read_json_manifest(manifest_path)
 
 
-def dump_manifest(contents: t.Dict, manifest_path: t.Union[Path, str]):
+def dump_manifest(
+    contents: t.Dict, manifest_path: t.Union[Path, str], has_yaml_header: bool = False
+):
     """Writes back 'contents' to 'manifest_path'.
 
     For YAML, we make a best-effort attempt to preserve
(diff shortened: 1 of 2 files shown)


src/lib/utils.py
```diff
@@ -512,7 +512,10 @@ def read_json_manifest(manifest_path: Path):
 def read_yaml_manifest(manifest_path: Path):
     """Read a YAML manifest from 'manifest_path'."""
     with manifest_path.open("r") as f:
-        return _yaml.load(f)
+        has_yaml_header = (_ := f.readline().strip()) == "---"
```
wjt: You can put a comment on the document separator line:

```yaml
--- # what a nice document
foo: bar
```

Annoyingly `ruamel.yaml` can't quite round-trip this:

```python
>>> yaml = YAML()
>>> yaml.explicit_start = True
>>> yaml.dump(yaml.load("""--- # what a nice document\nfoo: bar"""), sys.stdout)
---
    # what a nice document
foo: bar
```

But given that I think you could relax this check a little:


```suggestion
        has_yaml_header = (_ := f.read(3)) == "---"
```

src/lib/utils.py
```diff
@@ -562,6 +567,7 @@ def dump_manifest(contents: t.Dict, manifest_path: t.Union[Path, str]):
 
     with manifest_path.open("w", encoding="utf-8") as fp:
         if manifest_path.suffix in (".yaml", ".yml"):
+            _yaml.explicit_start = has_yaml_header  # type: ignore[assignment]
```
wjt, line 570: Why is this annotation needed? I guess I see above that `_yaml.width` needs the same thing.

wjt (approved):
Really annoying that there's no way to ask the YAML parser whether the document separator was necessary but having spent a while digging at ruamel.yaml (as I'm sure you did too) I agree that there's no better way than looking directly at the file. :(

A couple of suggestions you could take or leave:

src/lib/utils.py
bbhtt: It's the same issue as width, it thinks it is set to None. https://sourceforge.net/p/ruamel-yaml/tickets/322/

src/lib/utils.py
bbhtt: Makes sense.

PR #469, opened 2025-05-13
contributing: Remove black cmd from run-in-container.sh

bbhtt:


bbhtt:
hrm CI uses pypi to get black...

PR #474, opened 2025-07-09
gnomechecker: Allow major version 0 being stable

A6GibKm:
libdex has 0.10.0 as its latest stable release and the checker detects 0.11.0 as an update, but this version is unstable.

Updates from a major version of 0 to another major version =0 are not an issue, since the app was using such a major version in the first place.


A6GibKm:
cc @wjt

wjt (approved):
OK!

PR #475, opened 2025-07-26
manifest: Don't allow invalid appdata when change is a downgrade

### LINKED ISSUE #36: Generates invalid appdata when change is a downgrade
On https://github.com/flathub/us.zoom.Zoom/pull/35 the checker correctly submitted an update to a newer version of Zoom. The rotating URL was then rolled back to the older version, and the checker dutifully opened https://github.com/flathub/us.zoom.Zoom/pull/36. However, the latter PR prepended the new version to the appdata `<releases>`:

```patch
   <releases>
+    <release version="3.0.287250.0828" date="2019-08-29"/>
     <release version="3.0.291715.0908" date="2019-09-09"/>
     <release version="3.0.287250.0828" date="2019-08-29"/>
```

This is invalid:

```
Processing application us.zoom.Zoom
us.zoom.Zoom.desktop:              AppData problem: tag-invalid : <release> version was duplicated
us.zoom.Zoom.desktop:              AppData problem: tag-invalid : <release> versions are not in order [3.0.287250.0828 before 3.0.291715.0908]
```

Two changes suggest themselves:

* Validate the generated appdata with `appstream-util` before sending the PR
* Avoid generating this invalid change in the first place

The second could be done in a couple of ways:

* The cheat's way to do it is to only include one `<release>` in `<releases>` and just update it every time the version changes. However, the Zoom appdata has been dutifully preserving the full release history since before this checker entered the scene.
* The better approach would be to drop all `<release>` tags that do not compare older than the one we are about to prepend.

bbhtt:
Instead restore the original appdata back

Fixes https://github.com/flathub-infra/flatpak-external-data-checker/issues/36

### DIFF
--- a/dependencies.apt.txt
+++ b/dependencies.apt.txt
@@ -1,4 +1,5 @@
 bubblewrap
+gir1.2-appstream-1.0
 gir1.2-glib-2.0
 gir1.2-json-1.0
 git
--- a/src/lib/utils.py
+++ b/src/lib/utils.py
@@ -50,8 +50,10 @@ from .checksums import MultiHash
 
 import gi
 
+gi.require_version("AppStream", "1.0")
+gi.require_version("Gio", "2.0")
 gi.require_version("Json", "1.0")
-from gi.repository import GLib, Json  # noqa: E402
+from gi.repository import AppStream, Gio, GLib, Json  # noqa: E402
 
 log = logging.getLogger(__name__)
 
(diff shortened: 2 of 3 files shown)


bbhtt:
Draft because like it or not, generating broken appdata makes linter fail, thereby preventing merges.

and also this does not fix the core issue - downgrades should not occur in the first place.

bbhtt:
to prevent downgrades, i've been thinking of using `AppStream.vercmp_test_match()`...

PR #482, opened 2025-10-10
increase the maximum size of the manifest file

aleixpol:
The org.kde.Sdk manifest file is bigger (110 KiB) and this makes it not work. This change increases it to 1MiB.

### DIFF
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -53,7 +53,7 @@ from .checkers import Checker, ALL_CHECKERS
 
 MAIN_SRC_PROP = "is-main-source"
 IMPORTANT_SRC_PROP = "is-important"
-MAX_MANIFEST_SIZE = 1024 * 100
+MAX_MANIFEST_SIZE = 1024 * 1024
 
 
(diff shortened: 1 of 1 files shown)


bbhtt:
1 MB default is ok these days I think, this seems was set for performance concerns but I'm not sure if that's too much problem, parsing super large jsons shouldn't be a problem

Also note that you can use the CLI arg to set any manifest size you want.

bbhtt (approved):

bbhtt:
Hmm looks like I was wrong, it was done to prevent changes in generated manifests like the ones generated by cargo or node generator https://github.com/flathub-infra/flatpak-external-data-checker/issues/111

But using the size as a heuristic for that seems off.

bbhtt:
Seems like it shouldn't be a problem since https://github.com/flathub-infra/flatpak-external-data-checker/commit/c52d5e01ce2f790b83d864bb5fff6b9c595e5110, none of those large generated manifests generate x-checker data for them, other than pip generator, but that's intentional.

PR #483, opened 2025-10-10
tests: Switch to ftpmirror.gnu.org

bbhtt:


bbhtt:
Looks like coveralis is gone now.

PR #485, opened 2025-10-13
main: Support committing directly to the base branch

### LINKED ISSUE #255: Option to commit to the current branch
Sometimes a separate update branch may be unwanted, e.g. if the update isn't going through PR workflow, or the workflow needs to be customized. In such case, an option to commit directly to the current branch can be useful.

bbhtt:
Fixes https://github.com/flathub-infra/flatpak-external-data-checker/issues/255

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -172,7 +172,7 @@ def ensure_git_safe_directory(checkout: Path):
 class CommittedChanges(t.NamedTuple):
     subject: str
     body: t.Optional[str]
-    commit: str
+    commit: t.Optional[str]
     branch: str
     base_branch: t.Optional[str]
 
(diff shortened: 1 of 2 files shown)


src/main.py
```diff
@@ -225,9 +239,26 @@ def commit_changes(changes: t.List[str]) -> CommittedChanges:
...
 
-    # Moved to detached HEAD
-    log.info("Switching to detached HEAD")
-    check_call(["git", "-c", "advice.detachedHead=false", "checkout", "HEAD@{0}"])
+    if target_branch:
+        log.info("Committing directly to existing branch '%s'", target_branch)
+        if not branch_exists(target_branch):
+            log.error("Branch '%s' does not exist locally", target_branch)
+            return None
+        check_call(["git", "checkout", target_branch])
```
wjt: Running `git checkout` at this point may fail. By this point there are uncommitted changes in the working copy. Suppose you are on branch `main`, and you say `--commit-branch test`. If `test` exists but there are divergent changes that conflict with the uncommitted changes, then `git checkout test` will fail.

I think it would be simpler for this option to not take an argument. Instead you would say `--dont-switch-branch` (name tbd) and it would commit directly to whatever branch happens to be checked out, and not create a new one.

bbhtt: Yea, seems fine we can do `--commit-to-base-branch`.

bbhtt:
maybe --commit-to-current-branch instead, i don't want to give the impression that it'll find the base from random checked out commits etc.

PR #498, opened 2026-01-29
gnome: Do not crash if minor is not numeric

A6GibKm:
when the odd-minor-is-unstable version scheme is used. Instead we simply say it is not stable, the minor should be numerical with this scheme.

This fixes a crash when checking libgweather versions which uses this scheme but has a 40.alpha release.


A6GibKm:
@bbhtt thanks for reviewing. https://github.com/flathub/org.gnome.Maps/pull/148 already uses this, but the tool at Flathub has not yet updated, I fear it might be crashing whenever it checks if there are  updated.

bbhtt:
Flathub uses the docker image which is already updated. I will update the Flatpak as well.

PR #509, opened 2026-04-23
tests: Reformat with black

bbhtt:


bbhtt:
it's annoying that black keeps changing the formatting between versions so often sigh...

PR #511, opened 2026-04-23
utils: Use asyncio.gather with fail-fast behaviour

### LINKED ISSUE #420: Hangs (sometimes) if a checker raises an assertion failure
In #418, @tobiasmicheler figured out that the checker was hanging while checking https://github.com/flathub/com.modrinth.ModrinthApp/blob/1d954cc7c91841d020fe4c860593789a16ca62c6/com.modrinth.ModrinthApp.yml#L108. This is because prior to 1a0e0c76449a20344584644cca69c6c61d494af2, `gitchecker.GitChecker._check_has_new` asserted that the (user-supplied) regular expression contained exactly one match group.

Stylistically I think it was incorrect to make an assertion about user-supplied data: assertions are to catch logic errors in the checker itself. But nonetheless it is the case that assertions are used liberally in the various checkers, so it's a problem that the asyncio mainloop would hang in this situation.

The symptom is that once the assertion is raised, the program hangs; if you interrupt it with `Ctrl+C` you get the following backtrace:

```
^CTraceback (most recent call last):
  File "/usr/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/sysroot/home/wjt/src/flathub-infra/flatpak-external-data-checker/src/main.py", line 465, in run_with_args
  File "/sysroot/home/wjt/src/flathub-infra/flatpak-external-data-checker/src/manifest.py", line 408, in check
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sysroot/home/wjt/src/flathub-infra/flatpak-external-data-checker/src/manifest.py", line 343, in _check_data
    await checker.check(data)
  File "/sysroot/home/wjt/src/flathub-infra/flatpak-external-data-checker/src/checkers/gitchecker.py", line 92, in check
    return await self._check_has_new(external_data)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sysroot/home/wjt/src/flathub-infra/flatpak-external-data-checker/src/checkers/gitchecker.py", line 101, in _check_has_new
    assert tag_re.groups == 1
           ^^^^^^^^^^^^^^^^^^
AssertionError

During handling of the above exception, another exception occurred:

  File "/sysroot/home/wjt/src/flathub-infra/flatpak-external-data-checker/./flatpak-external-data-checker", line 30, in <module>
    main()
  File "/sysroot/home/wjt/src/flathub-infra/flatpak-external-data-checker/src/main.py", line 504, in main
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/runners.py", line 189, in run
    with Runner(debug=debug) as runner:
  File "/usr/lib/python3.11/asyncio/runners.py", line 63, in __exit__
    self.close()
  File "/usr/lib/python3.11/asyncio/runners.py", line 71, in close
    _cancel_all_tasks(loop)
  File "/usr/lib/python3.11/asyncio/runners.py", line 201, in _cancel_all_tasks
    loop.run_until_complete(tasks.gather(*to_cancel, return_exceptions=True))
  File "/usr/lib/python3.11/asyncio/base_events.py", line 640, in run_until_complete
    self.run_forever()
  File "/usr/lib/python3.11/asyncio/base_events.py", line 607, in run_forever
    self._run_once()
  File "/usr/lib/python3.11/asyncio/base_events.py", line 1884, in _run_once
    event_list = self._selector.select(timeout)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/selectors.py", line 468, in select
    fd_event_list = self._selector.poll(timeout, max_ev)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyboardInterrupt
```

A bit of printf debugging shows that the gitchecker for another module is executing the async function `git_ls_remote` at the point of the hang. If I remove that module, the assertion bubbles up to the top level correctly.

bbhtt:
asyncio.gather does not cancel sibling tasks when one fails, so if one
checker raises an unhandled exception, other tasks running I/O may keep
the event loop alive indefinitely. So use asyncio_gather_failfast to
detect failures and explicitly cancel+drain all pending tasks before
re-raising the exception.

In Python 3.11+ this is offered by TaskGroup but it is cheap to
use a wrapper here.

See https://github.com/python/cpython/issues/75633

Fixes: https://github.com/flathub-infra/flatpak-external-data-checker/issues/420

### DIFF
--- a/src/lib/utils.py
+++ b/src/lib/utils.py
@@ -57,6 +57,8 @@ from gi.repository import (  # type: ignore[attr-defined] # noqa: E402 # noqa: I
     Json,
 )
 
+T = t.TypeVar("T")
+
 log = logging.getLogger(__name__)
 
 
@@ -628,3 +630,29 @@ def init_logging(level=logging.DEBUG):
     logging.basicConfig(level=level, format="%(levelname)-7s %(name)s: %(message)s")
     if level == logging.DEBUG:
         logging.getLogger("github.Requester").setLevel(logging.INFO)
+
+
+async def asyncio_gather_failfast(
+    awaitables: t.Iterable[t.Awaitable[T]],
+) -> list[T]:
+    tasks: list[asyncio.Task[T]] = [asyncio.ensure_future(a) for a in awaitables]
+    if not tasks:
+        return []
+    done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_EXCEPTION)
+    exc: BaseException | None = None
+    for task in done:
+        if task.cancelled():
+            continue
+        try:
+            e = task.exception()
+        except asyncio.CancelledError:
+            continue
+        if e is not None:
+            exc = e
+            break
+    if exc:
+        for task in pending:
+            task.cancel()
+        await asyncio.gather(*pending, return_exceptions=True)
+        raise exc
+    return [task.result() for task in tasks]
--- a/src/manifest.py
+++ b/src/manifest.py
@@ -18,7 +18,6 @@
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
-import asyncio
 import dataclasses
 import datetime
 import logging
@@ -48,7 +47,7 @@ from .lib.externaldata import (
     BuilderModule,
     ExternalBase,
 )
-from .lib.utils import dump_manifest, read_manifest
+from .lib.utils import asyncio_gather_failfast, dump_manifest, read_manifest
 
 MAIN_SRC_PROP = "is-main-source"
 IMPORTANT_SRC_PROP = "is-important"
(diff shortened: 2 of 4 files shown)


bbhtt:
This seems correct to me but I didn't manage to reproduce the hang with 1a0e0c76449a20344584644cca69c6c61d494af2 reverted and a broken manifest locally. I think it requires some running / semi-stuck network I/O and I can't get that. So it's a bit theoretical.

PR #512, opened 2026-04-23
Add support for robots.txt

### LINKED ISSUE #434: Check and respect robots.txt
I've just encountered a project that responds 403 Forbidden to requests with FEDC's User-Agent. Once I realised this was what was happening, I experimented with a few different user-agents for a few minutes, and then my IP address was blocklisted by the web server, which no longer responds to TCP connection attempts.

Searching the project's forum (via a VPN!) I came across [this thread](https://forum.filezilla-project.org/viewtopic.php?f=1&t=55764&p=183884&hilit=403#p183880) where the author refers to:

> automated third-party software in your network that disregards robots.txt or HTTP 403.

I checked the project's robots.txt and there is nothing there that the tools I was using (fedc, wget, and Debian's uscan) were disobeying. But it's true that FEDC does not check robots.txt. It probably should, with a per-domain cache that would ideally persist between runs of the tool.

bbhtt:
Fixes https://github.com/flathub-infra/flatpak-external-data-checker/issues/434


bbhtt:
I need to do some digging and figure out which checkers to enable/disable this for and probably add some argument to override.

```
→ curl -s https://api.snapcraft.io/robots.txt                                                                                                                   
User-agent: *
Disallow: /
```

bbhtt:
GitHub, Gitlab also blocks necessary paths commonly used for all U/As. PyPI chromium, rust, snapcraft blocks things entirely

I will make this opt-in on per source and gloabl basis instead.

bbhtt:
Left the boilerplate mock patches the checker tests in case later we enable...

bbhtt:
can I get reviews for these? cc @wjt

wjt:
The code looks fine. It is a bit weird to implement this but then ignore robots.txt because half of the APIs we hit block the API in robots.txt. (e.g. https://pypi.org/robots.txt matches all the /json endpoints documented in https://docs.pypi.org/api/json/ - what's the point of an API if not for a robot? Maybe fedc is not considered a robot in this case?)

bbhtt:
Not quite sure but it does make the feature a bit useless.

PR #524, opened 2026-06-11
Add minimal pyproject.toml and use uv & ruff

bbhtt:
There's a million different errors from ruff and mypy as I enabled strict mode. Those will need fixing one by one, otherwise this should be the mostly ready setup.

```
This also switches to ruff for formatting and linting replacing
black and flake8

The benefits are:

- CI can be greatly simplified
- Easy to set up the linting and formatting env locally
- Upgrading / managing dev dependencies becomes easier
- uv allows to set up venvs with selected python versions easily
  which is perfect for a project using some aging dependencies
  and running / being tested in a very selected slow moving
  environment such as Debian stable

Benefits of ruff:

- Contains a vast number of rules and lints covering pylint, flake8
  etc.
- Unifies formatting and linting workflow
```


bbhtt:
It would have nice to use the full checks of mypy strict mode, it's what I use elsewhere, it does get a verbose but it helps in the long run. but there's a lot of code here and changing everything here is not a good idea. I ended up disabling no-untyped-def, no-untyped-call and type-arg.

bbhtt:
The problem with not type hinting all definitions is later we have to narrow types with assert,cast etc, which I find is "dumb", and when it is ignored, people will spam those instead of adding the actual fix - adding the type hints. But it is what it is...

Would be nice to have a review for this soon, otherwise it will infinitely conflict with everything.

PR #528, opened 2026-06-13
anityachecker: Propagate latest_version_created_on as timestamp

### LINKED ISSUE #507: Add timestamp-query support to AnityaChecker when upstream exposes the information in API
I'm encountering #477 on https://github.com/flathub/io.github.go_musicfox.go-musicfox because the Anitya checker is missing the `timestamp-query` argument to specify using the release date as the timestamp.

Anitya seems to expose this information on their site, so amending the API should be possbile.

Or perhaps the best course of action is to migrate from Anitya to the JSON checker?

bbhtt:
Fixes: https://github.com/flathub-infra/flatpak-external-data-checker/issues/507

### DIFF
--- a/src/checkers/anityachecker.py
+++ b/src/checkers/anityachecker.py
@@ -1,4 +1,6 @@
 import logging
+import re
+from datetime import datetime, timezone
 
 from yarl import URL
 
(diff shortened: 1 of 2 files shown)


bbhtt:
Managed to test this today.

